PR #57, opened 2024-01-16
Performance related changes

nt-gt:
These changes remove a considerable pressure on the CPU time spent, meaning the CPU is no longer spinning close to 100% when the eBL conformance tests are running (it is down to ~40% for me now).

While it makes things a bit faster, it does not translate into a 1:1 improvement in wall time.  The profile now only sees about 1 minute of runtime but wall time for the full eBL test still takes over 4 minutes. Even with the remaining CPU time left, half of it now appears directly in the java TCP Connection manager stack and non in CTK code.


nt-gt:
Note: This change removes the tracking of `max state length` (as mentioned in an email that was causing me grief) on the assumption that is not "that" important.

PR #61, opened 2024-01-24
DT-380

preetamnpr:


booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
```diff
@@ -381,6 +381,7 @@ private void generateTransportPlan(ObjectNode booking) {
 
   private void replaceShipmentCutOffTimes(ObjectNode booking) {
     var shipmentCutOffTimes = booking.putArray("shipmentCutOffTimes");
+    var receiptTypeAtOrigin = booking.get("receiptTypeAtOrigin").asText(null);
```
nt-gt, line 384: Throws an NPE if that field is omitted.  Please use `.path(...)` instead of `.get(...)`

nt-gt: Also, the `.asText(null)` could have been `.asText("")` so `receiptTypeAtOrigin` could never be null (there is no value in allowing `null` here in the first place, so there is only risks by allowing)

PR #64, opened 2024-01-31
DT-853

preetamnpr:


booking/src/main/java/org/dcsa/conformance/standards/booking/checks/ShipperBookingContentConformanceCheck.java
```diff
@@ -39,61 +41,63 @@ private boolean isReeferContainerSizeTypeCode(String isoEquipmentCode) {
...
-            );
-          } else {
-            fieldOmitted(
-              requestedEquipment,
-              ACTIVE_REEFER_SETTINGS_FIELD,
-              issues,
-              prefix,
-              "the isNonOperatingReefer was true"
-            );
+    if (!scenarioType.equals(ScenarioType.REGULAR_NON_OPERATING_REEFER)) {
```
nt-gt, line 45: I do not understand why you are making this validation conditional on the scenario. The validation here was not intended to be scenario specific.

booking/src/main/java/org/dcsa/conformance/standards/booking/party/BookingCarrier.java
```diff
@@ -105,16 +103,72 @@ private String generateSchemaValidVesselIMONumber() {
...
-        new CarrierScenarioParameters(
-            "Carrier Service %d".formatted(RANDOM.nextInt(999999)),
-            generateSchemaValidVesselIMONumber(),
-            "service name",
-            validHsCodeAndCommodityType.get(0),
-            validHsCodeAndCommodityType.get(1),
-            generateValidPolUNLocationCode(),
-            generateValidPodUNLocationCode());
+    CarrierScenarioParameters carrierScenarioParameters = switch (scenarioType) {
+    case REGULAR, REGULAR_SHIPPER_OWNED -> new CarrierScenarioParameters("Example Carrier Service",
```
nt-gt, line 107: I prefer the previous way of providing a service name other this static `Example Carrier Service` string (applies to more lines than I highlighted)

nt-gt (changes requested):
I am missing a validation to ensure the shipper provides the current scenario parameters.

As an example, for the `REGULAR_NON_OPERATING_REEFER` scenario, the shipper must supply `isNonOperatingReefer = false`.

I am aware that it was not present prior to your changes, but there is no point in all of these scenarios of the shipper can get a conform respond without actually following the scenario.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/ShipperBookingContentConformanceCheck.java
preetamnpr: I have created a story to add the scenario based checks DT-897

nt-gt (approved):

PR #66, opened 2024-02-08
DT-898 Defaulting to binary HTTP body payloads

gj0dcsa:


core/src/main/java/org/dcsa/conformance/core/traffic/ConformanceMessageBody.java
```diff
@@ -1,22 +1,45 @@
...
+  private final boolean isRawByteArray;
+  private final byte[] byteArrayBody;
   private final String stringBody;
   private final JsonNode jsonBody;
+  private final boolean isCorrectJson;
+
+  public ConformanceMessageBody(byte[] byteArrayBody) {
+    this.isRawByteArray = true;
+    this.byteArrayBody = byteArrayBody == null ? new byte[0] : byteArrayBody;
+    this.stringBody = new String(this.byteArrayBody, StandardCharsets.UTF_8);
```
nt-gt, line 25: How does this work if the content is not valid UTF-8 (which is likely for the use-cases that need raw byte transfers)?

nt-gt: I specifically need support for https://app.swaggerhub.com/apis/dcsaorg/DCSA_EBL_PINT/3.0.0-Beta-1#/Transfer%20additional%20document/put_v3_envelopes__envelopeReference__additional_documents__documentChecksum_, which is about transferring additional documentation.

Said documentation can be **any** file including but not limited to PDF and will be transferred in the application/octet-stream stream. The validation will need to assert this stream matches the expected checksum, which is why we need binary transfers. However, there is no rule stating that the content will be a valid UTF-8 byte sequence. In fact, for a PDF, a zip file, scanned images, it will very likely **not** be valid UTF-8.

PR #67, opened 2024-02-12
DT-876: Support basic scenario without additional documentation

nt-gt:
Also covers DT-873 and DT-874 as a side-effect.

### DIFF
--- a/pint/LICENSE
+++ b/pint/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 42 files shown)


pint/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
preetamnpr, line 189: Year and name is missing. is this yet to be defined ? If so, please create the jira to decide and update this file

nt-gt: This is part of the standard license and not a placeholder we have to fill out.

preetamnpr (approved):
looks fine to me.

PR #71, opened 2024-02-28
Dt 786

preetamnpr:


booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -598,9 +602,9 @@ private static void generateScenarioRelatedChecks(List<JsonContentCheck> checks,
         var charges = body.path("charges");
         var issues = new LinkedHashSet<String>();
         for(JsonNode charge : charges) {
-          var chargeAmount = charge.get("chargeAmount").asDouble();
-          if (BigDecimal.valueOf(chargeAmount).scale() > 2) {
-            issues.add("Charge amount %s is expected to have 2 decimal precious ".formatted(chargeAmount));
+          var currencyAmount = charge.get("currencyAmount").asDouble();
```
nt-gt, line 605: Use `.path` instead of `.get` to avoid NPE

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-reefer-temp-change.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REEFER-TEMP",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-dg.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-DG",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-reefer.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REEFER",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-2re1c.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-2RE1C",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-2re2c.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-2RE2C",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-cho-dest.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "SD",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-CHO-DEST",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-cho-orig.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-CHO-ORIG",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-non-operating-reefer.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-NON-OP-REEFER",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular-shipper-owned.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR-SHIPPER-OWNED",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-Beta-1-regular.json
```diff
@@ -3,6 +3,7 @@
   "deliveryTypeAtDestination": "CY",
   "cargoMovementTypeAtOrigin": "FCL",
   "cargoMovementTypeAtDestination": "FCL",
+  "serviceContractReference": "SCR-1234-REGULAR",
```
nt-gt, line 6: This should probably a parameter that the carrier can choose.

nt-gt (changes requested):
Added some more remarks.

In addition, I do not see a fix for the remark I made inhttps://github.com/dcsaorg/Conformance-Gateway/pull/71#pullrequestreview-1905776565

preetamnpr:
> Added some more remarks.
> 
> In addition, I do not see a fix for the remark I made inhttps://github.com/[/pull/71](https://github.com/dcsaorg/Conformance-Gateway/pull/71)#pullrequestreview-1905776565

I understood now what you actually mean.. I will put these checks across different use cases.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -294,43 +294,6 @@ private static Consumer<MultiAttributeValidator> allDg(Consumer<MultiAttributeVa
     }
   );
 
-
```
preetamnpr: please note this validation is nor more valid in booking and hence removed after confirmation from Flavia and Henrik

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -461,23 +424,44 @@ private static <T, O> Supplier<T> delayedValue(Supplier<O> cspSupplier, Function
...
       "[Scenario] Verify that the correct 'carrierServiceName' is used",
-      "carrierServiceName",
-      delayedValue(cspSupplier, CarrierScenarioParameters::carrierServiceName)
+      (body) -> {
+        var carrierServiceName = body.path("carrierServiceName");
+        var issues = new LinkedHashSet<String>();
+        if(carrierServiceName != null) {
+          delayedValue(cspSupplier, CarrierScenarioParameters::carrierServiceName  );
+        }
+        return issues;
```
nt-gt, line 437: This code does not ever issue an error for non-conformance. Accordingly, there must be a bug somewhere.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -461,23 +424,44 @@ private static <T, O> Supplier<T> delayedValue(Supplier<O> cspSupplier, Function
...
-      "contractQuotationReference",
-      delayedValue(cspSupplier, CarrierScenarioParameters::contractQuotationReference)
+      (body) -> {
+        var contractQuotationReference = body.path("contractQuotationReference");
+        var issues = new LinkedHashSet<String>();
+        if(contractQuotationReference != null) {
+          delayedValue(cspSupplier, CarrierScenarioParameters::contractQuotationReference);
+        }
+        return issues;
+      }
```
nt-gt, line 450: Ditto, there is no way to get a non-conformant result

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -461,23 +424,44 @@ private static <T, O> Supplier<T> delayedValue(Supplier<O> cspSupplier, Function
...
-      "carrierExportVoyageNumber",
-      delayedValue(cspSupplier, CarrierScenarioParameters::carrierExportVoyageNumber)
+      (body) -> {
+        var carrierExportVoyageNumber = body.path("carrierExportVoyageNumber");
+        var issues = new LinkedHashSet<String>();
+        if(carrierExportVoyageNumber != null) {
+          delayedValue(cspSupplier, CarrierScenarioParameters::carrierExportVoyageNumber);
+        }
+        return issues;
+      }
```
nt-gt, line 462: Ditto, there is no way to get a non-conformant result.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: Causing the check to never fail is not the answer. It just turns the code into dead-weight and a false sense of security.

Instead, why are you doing validation checks on a use-case where the validation check is not relevant? The scenario checks applies to all use-cases where you can check the scenario decided values, not blindly to all use-cases.

preetamnpr: I agree to this..

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: Yeah looks to be the case. I will reinstate it.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+      var carrierExportVoyageNumber = body.path("carrierExportVoyageNumber");
+      var issues = new LinkedHashSet<String>();
+
+      if (carrierExportVoyageNumber == null || (expectedDepartureDate == null || ( expectedArrivalEndDate == null && expectedArrivalStartDate == null)))  {
+        issues.add("The carrierExportVoyageNumber must be present as either expectedDepartureDate or expectedArrivalDates are not present");
+      }
+
+      return issues;
+    }
+  );
```
nt-gt, line 190: Please provide an `atLeastOneOf` method in the `JsonAttribute` (you can use `atMostOneOf` from the latest dev branch as a starting point) and rewrite this check to use `atLeastOneOf`.

nt-gt: Nvm, this is not a true "at least one of"

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+      var receiptTypeAtOrigin = body.path("receiptTypeAtOrigin").asText("");
+      var issues = new LinkedHashSet<String>();
+      var cutOffDateTimeCodes = StreamSupport.stream(shipmentCutOffTimes.spliterator(), false)
+        .map(p -> p.path("cutOffDateTimeCode"))
+        .filter(JsonNode::isTextual)
+        .map(n -> n.asText(""))
+        .collect(Collectors.toSet());
+      if (receiptTypeAtOrigin.equals("CFS")) {
+        if (!cutOffDateTimeCodes.contains("LCL")) {
+          issues.add("For ReceiptTypeAtOrigin as CFS, the shipment cut off date time must have only LCL");
```
nt-gt, line 284: Message says `only` but the requirement does not prevent other `shipment cut off date time` codes.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+        }
+      }
+      else {
+        if (cutOffDateTimeCodes.stream().noneMatch(CUTOFF_DATE_TIME_CODES::contains)) {
+          issues.add("The cutOff times doesn't have the allowed values" + CUTOFF_DATE_TIME_CODES);
+
+        }
```
nt-gt, line 291: Extract this to a separate check that uses `JsonAttribute.allIndividualMatchesMustBeValid` with `JsonAttribute.matchedMustBeDatasetKeywordIfPresent`

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+
+
+  private static final JsonContentCheck REQUESTED_CHANGES_PRESENCE = JsonAttribute.customValidator(
+    "Requested changes must be present for the selected Booking Status ",
+    (body) -> {
+      var bookingStatus = body.path("bookingStatus").asText("");
+      var issues = new LinkedHashSet<String>();
+      if (PENDING_CHANGES_STATES.contains(BookingState.fromWireName(bookingStatus))) {
+        var requestedChanges = body.path("requestedChanges");
+        if (requestedChanges == null) {
```
nt-gt, line 363: `requestedChanges` cannot be `null` (`.path` deliberately never returns null)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+    }
+  );
+
+  private static final JsonContentCheck REASON_PRESENCE = JsonAttribute.customValidator(
+    "Requested changes must be present for the selected Booking Status ",
+    (body) -> {
+      var bookingStatus = body.path("bookingStatus").asText("");
+      var issues = new LinkedHashSet<String>();
+      if (REASON_STATES.contains(BookingState.fromWireName(bookingStatus))) {
+        var reason = body.path("reason");
```
nt-gt, line 377: `reason` cannot be null (`.path` deliberately never returns null)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+        if (body.path("charges") == null) {
+          issues.add("charges for confirmed booking is not present");
+        }
+        if (body.path("shipmentCutOffTimes") == null) {
+          issues.add("shipmentCutOffTimes for confirmed booking is not present");
+        }
+        if (body.path("advanceManifestFilings") == null) {
+          issues.add("advanceManifestFilings for confirmed booking is not present");
+        }
+      }
```
nt-gt, line 413: The `.path` cannot return null, which makes many of these `if` statements always be false.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,659 @@
...
+      String expectedDepartureDate = body.path("expectedDepartureDate").asText("");
+      var invalidDates = new LinkedHashSet<String>();
+      if(!expectedDepartureDate.isEmpty()) {
+        LocalDate expectedDepartureLocalDate = LocalDate.parse(expectedDepartureDate);
+        if (expectedDepartureLocalDate.isBefore(LocalDate.now())) {
+          invalidDates.add(expectedDepartureLocalDate.toString());
+        }
+      }
+      return invalidDates.stream()
+        .map("The expected departure date '%s' can not be paste date"::formatted)
```
nt-gt, line 89: Spelling

booking/src/main/java/org/dcsa/conformance/standards/booking/action/Shipper_GetBookingAction.java
```diff
@@ -82,46 +80,7 @@ protected Stream<? extends ConformanceCheck> createSubChecks() {
                 getMatchedExchangeUuid(),
                 HttpMessageType.RESPONSE,
                 responseSchemaValidator),
-            new CarrierGetBookingPayloadResponseConformanceCheck(
```
nt-gt, line 85: Is this class still used ? If not, please delete it.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,654 @@
...
+        var dg = nodeToValidate.path("dangerousGoods");
+        if (!dg.isArray() || dg.isEmpty()) {
+          return Set.of();
+        }
+        if (nodeToValidate.path("packageCode").isMissingNode() && nodeToValidate.path("imoPackagingCode").isMissingNode()) {
+          return Set.of("The '%s' object did not have a 'packageCode' nor an 'imoPackagingCode', which is required due to dangerousGoods"
+            .formatted(contextPath));
+        }
+        if (nodeToValidate.path("numberOfPackages").isMissingNode() && nodeToValidate.path("description").isMissingNode()) {
+          return Set.of("The '%s' object did not have a 'numberOfPackages' nor an 'description', which is required due to dangerousGoods"
```
nt-gt, line 547: This issue does not match the title of the check.  Consider splitting the check into two.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,654 @@
...
+    UNIVERSAL_SERVICE_REFERENCE,
+    VALIDATE_SHIPMENT_CUTOFF_TIME_CODE,
+    VALIDATE_ALLOWED_SHIPMENT_CUTOFF_CODE,
+    VALIDATE_DOCUMENT_PARTIES,
+    JsonAttribute.atLeastOneOf(
+      JsonPointer.compile("/expectedDepartureDate"),
+      JsonPointer.compile("/expectedArrivalAtPlaceOfDeliveryStartDate"),
+      JsonPointer.compile("/expectedArrivalAtPlaceOfDeliveryEndDate"),
+      JsonPointer.compile("/carrierExportVoyageNumber")
+    ),
```
nt-gt, line 533: I see you changed this into a `atLeastOneOf`. Did you clarify with Flavia whether this is truly a `atLeastOneOf` situation?  (I think the earlier version of code I commented on turned out not to be a simple `one of` and I had to amend my remark when I re-read the code)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,654 @@
...
+    return JsonAttribute.contentChecks(
+      BookingRole::isCarrier,
+      matched,
+      HttpMessageType.RESPONSE,
+      checks
+    );
+  }
+
+
+  private static JsonContentMatchedValidation combineAndValidateAgainstDataset(
```
nt-gt, line 629: Rather than copy-wasting it into `BookingChecks` it would probably make sense to make a `public static` version of this in `JsonAttribute` and then remove the code duplication from the EblChecks.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,654 @@
...
+        for(JsonNode charge : charges) {
+          var currencyAmount = charge.path("currencyAmount").asDouble();
+          if (BigDecimal.valueOf(currencyAmount).scale() > 2) {
+            issues.add("Charge amount %s is expected to have 2 decimal precious ".formatted(currencyAmount));
+          }
+        }
+        return issues;
+      }
+    )
+    );
```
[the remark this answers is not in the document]
nt-gt: Then you would also get a proper path for the issue, so people can trivially find where in the JSON the problem is.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: (also, if the split is not done, then both issues should be emittable at the same time as they are not dependent on each other)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,654 @@
...
+      BOOKING_STATUS,
+      bookingStatus.wireName()
+    ));
+    checks.add(ADVANCED_MANIFEST_FILING_CODES_UNIQUE);
+    checks.add(AMF_CC_MTC_COMBINATION_VALIDATIONS);
+    checks.add(SHIPMENT_CUTOFF_TIMES_UNIQUE);
+    checks.add(CHECK_CONFIRMED_BOOKING_FIELDS);
+    checks.add(VALIDATE_SHIPMENT_LOCATIONS);
+    checks.add(REQUESTED_CHANGES_PRESENCE);
+    checks.add(REASON_PRESENCE);
```
nt-gt, line 609: These looks like they ought to be in a static final list of some form with a `checks.addAll(...)`.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: I checked with Henrik and he confirmed that atleast any of these fields must be present.

booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
```diff
@@ -487,6 +493,27 @@ private void addCharge(ObjectNode booking) {
...
+      advanceManifestFilings
+        .addObject()
+        .put("manifestTypeCode", "ACI")
+        .put("countryCode", "EG");
+    } else {
+      advanceManifestFilings
+        .addObject()
+        .put("manifestTypeCode", "ACE")
+        .put("countryCode", "US");
+    }
```
nt-gt, line 513: This will still cause duplicates if the `addAdvancedManifestFilings` is called multiple for the same booking (we are now at thrice rather than twice, but the underlying bug remains).

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,651 @@
...
+      return issues;
+    }
+  );
+
+  private static final JsonContentCheck REASON_PRESENCE = JsonAttribute.customValidator(
+    "Reason field must be present for the selected Booking Status",
+    (body) -> {
+      var bookingStatus = body.path("bookingStatus").asText("");
+      var issues = new LinkedHashSet<String>();
+      if (REASON_STATES.contains(BookingState.fromWireName(bookingStatus))) {
```
nt-gt, line 356: This is the only usage of `REASON_STATES`. In this usage, `REASON_STATES` is only checking the `bookingStatus` attribute. However, `REASON_STATES` includes two status codes only applicable to `amendedBookingStatus`, which suggests that the logic here is not correct.

Note: You might have copied/moved this code from else where were the code might have been correct at the time it was written. However, what matters is whether the code is correct now and I suspect it is not.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/AbstractCarrierPayloadConformanceCheck.java
```diff
@@ -18,19 +18,12 @@ abstract class AbstractCarrierPayloadConformanceCheck extends PayloadContentConf
     BookingState.PENDING_AMENDMENT
   );
```
nt-gt, line 19: Is this class (and any of its subclasses) still used? Based on the changes, I would not be surprised if it was now redundant.

(If it is still used, leave it for now. Ideally, we should migrate all of them to the new BookingChecks method. However, this PR is large enough as it is, so I am fine with the move happening later if it was not already done).

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,651 @@
...
+    checks.addAll(STATIC_BOOKING_CHECKS);
+    checks.addAll(RESPONSE_ONLY_CHECKS);
+    if (CONFIRMED_BOOKING_STATES.contains(bookingStatus)) {
+      checks.add(COMMODITIES_SUBREFERENCE_UNIQUE);
+    }
+    if (BOOKING_STATES_WHERE_CBR_IS_OPTIONAL.contains(bookingStatus)) {
+      checks.add(JsonAttribute.mustBeAbsent(
+        CARRIER_BOOKING_REFERENCE
+      ));
+    }
```
nt-gt, line 635: This check is not correct. The carrier booking reference can be present in any state.  However, it **must** be present in some states.

Also, we should validate that the carrier booking reference is the same once it appears (just as we do with the carrier booking _request_ reference).

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,651 @@
...
+    JsonAttribute.allIndividualMatchesMustBeValid(
+      "The charges currency amount must not exceed more than 2 decimal points",
+      mav -> mav.submitAllMatching("charges.*"),
+      (nodeToValidate, contextPath) -> {
+        var currencyAmount = nodeToValidate.path("currencyAmount").asDouble();
+        if (BigDecimal.valueOf(currencyAmount).scale() > 2) {
+          return Set.of("currencyAmount %s is expected to have 2 decimal precious ".formatted(contextPath));
+        }
+        return Set.of();
+      }
```
nt-gt, line 602: This could be simplified even further by using `charges.*.currentAmount` rather than `charges.*` (would remove the need for the `.path` and the issue message could just use `%s` directly without having to clarify that it is `currentAmount` at `%s`, etc.)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: It looks like you attempted to split out of the change, but I still see the number of packages vs. description check embedded in the packagingCode/imoPackagingCode check

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,651 @@
...
+          issues.add("The 'SCO' party is mandatory when 'serviceContractReference' is absent");
+        }
+        if (!body.path("contractQuotationReference").isTextual()) {
+          issues.add("The 'SCO' party is mandatory when 'contractQuotationReference' is absent");
+        }
+      }
+      Set<String> partyShippers  = new HashSet<>(Arrays.asList("OS", "DDR"));
+      if (partyFunctions.stream().noneMatch(partyShippers::contains)) {
+        if (!partyFunctions.contains("BA")) {
+          issues.add("The 'BA' party must exist if 'OS' or 'DDR' party is absent");
```
nt-gt, line 497: Did you check this text with a native English speaker? I still feel that issue description is misleading relative to the logic applied.

preetamnpr: I used the online grammar tools to validate and seems to be grammatically fine.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,651 @@
...
+          .filter(o ->  o.path("locationTypeCode").asText("").equals("POL")
+            || o.path("locationTypeCode").asText("").equals("PDE") )
+          .findFirst()
+          .orElse(null);
+      var podNode =
+        StreamSupport.stream(shipmentLocations.spliterator(), false)
+          .filter(o -> o.path("locationTypeCode").asText("").equals("POD")
+          || o.path("locationTypeCode").asText("").equals("PRE") )
+          .findFirst()
+          .orElse(null);
```
nt-gt, line 314: I do not understand why `POL` and `PDE` are paired and ditto for `PRE` and `POD`.  Normally, the pairings are `PRE` + `POL` vs. `POD` + `PDE`.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: The sentence `The 'BA' party must exist if 'OS' or 'DDR' party is absent` and the sentence `The 'BA' party must exist if 'OS' and 'DDR' party are absent` are both grammatically valid sentences. However, they have a subtle and yet important difference in meaning.

The `BA` party is required when (both) `OS` **and** `DDR` are not present. Not when when (either one of) `OS` **or** `DDR` are absent. This is why I keep flagging this sentence.

booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
nt-gt: The example code would solve my concern indeed.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: missed to delete it.. thanks for pointing

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: shipment Cut-off Times / cut-off time code: 	LCL - only applicable when the Receipt Type at Origin is CFS.

nt-gt: Again, as I read the code and the issue text, they do not agree.

preetamnpr: "cutOffDateTimeCode 'LCL' is not present when receiptTypeAtOrigin is CFS"

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: This `mustBeAbsent` check is still here, but as stated it is incorrect. The standard never forbids the CBR in any state. The standard requires the CBR in certain states.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    BookingState.UPDATE_RECEIVED,
+    /* CANCELLED depends on whether cancel happens before CONFIRMED, but the logic does not track prior
+     * states. Therefore, we just assume it is optional in CANCELLED here.
+     */
+    BookingState.CANCELLED
+  );
+  private static final JsonPointer CARRIER_BOOKING_REQUEST_REFERENCE = JsonPointer.compile("/carrierBookingRequestReference");
+  private static final JsonPointer CARRIER_BOOKING_REFERENCE = JsonPointer.compile("/carrierBookingReference");
+  private static final JsonPointer BOOKING_STATUS = JsonPointer.compile("/bookingStatus");
+  private static final JsonPointer[] TD_UN_LOCATION_CODES = {
```
HenrikHL, line 61: If this is to list all places where a `UN Location Code` is needed - then you are missing the `PlaceOfBLIssue` plus `loadLocation` and `dischargeLocation` as part of the `Transport Plan`

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    ALL_AMF,
+    JsonAttribute.unique("countryCode", "manifestTypeCode")
+  );
+  private static final Consumer<MultiAttributeValidator> ALL_SHIPMENT_CUTOFF_TIMES = (mav) -> mav.submitAllMatching("shipmentCutOffTimes.*");
+
+  private static final JsonContentCheck SHIPMENT_CUTOFF_TIMES_UNIQUE = JsonAttribute.allIndividualMatchesMustBeValid(
+    "in 'shipmentCutOfftimes' cutOff date time code must be unique",
+    ALL_SHIPMENT_CUTOFF_TIMES,
+    JsonAttribute.unique("cutOffDateTimeCode")
+  );
```
nt-gt, line 288: Have you validated this one (as in deliberately try to make it fail to ensure it flags invalid input)? `unique` only works when used on an (unexpanded) array, but the validator seems to validate the `array`.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    }
+  );
+
+  private static final Consumer<MultiAttributeValidator> ALL_AMF = (mav) -> mav.submitAllMatching("advanceManifestFilings.*");
+
+  private static final JsonContentCheck ADVANCED_MANIFEST_FILING_CODES_UNIQUE = JsonAttribute.allIndividualMatchesMustBeValid(
+    "The combination of 'countryCode' and 'manifestTypeCode' in 'advanceManifestFilings' must be unique",
+    ALL_AMF,
+    JsonAttribute.unique("countryCode", "manifestTypeCode")
+  );
```
nt-gt, line 281: Have you validated this one (as in deliberately try to make it fail to ensure it flags invalid input)? unique only works when used on an (unexpanded) array, but the validator seems to validate the array.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt, line 61: Nit: The variable is named `TD_UN_LOCATION_CODES`, which I suspect is copy-waste from EBL. Since Booking does not have a transport document (TD), then name is misleading.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+     */
+    BookingState.CANCELLED
+  );
+  private static final JsonPointer CARRIER_BOOKING_REQUEST_REFERENCE = JsonPointer.compile("/carrierBookingRequestReference");
+  private static final JsonPointer CARRIER_BOOKING_REFERENCE = JsonPointer.compile("/carrierBookingReference");
+  private static final JsonPointer BOOKING_STATUS = JsonPointer.compile("/bookingStatus");
+  private static final JsonPointer[] TD_UN_LOCATION_CODES = {
+    JsonPointer.compile("/invoicePayableAt/UNLocationCode"),
+    JsonPointer.compile("/shipmentLocations/location/UNLocationCode"),
+    JsonPointer.compile("/shipmentLocations/location/facilityLocation/UNLocationCode")
```
nt-gt, line 64: This does not work; `shipmentLocations` is an array and JSON pointer can only match a single node. In the concrete case, the pointer matches nothing (since the path would have be /shipmentLocations/0/... + /shipmentLocations/1/... + ..., which will not scale).

You would have to use a different approach when arrays are involved.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    }
+  );
+
+  private static final JsonContentCheck CHECK_EXPECTED_ARRIVAL_POD = JsonAttribute.customValidator(
+    "Check expected arrival dates are valid",
+    (body) -> {
+      String providedArrivalStartDate = body.path("expectedArrivalAtPlaceOfDeliveryStartDate").asText("");
+      String providedArrivalEndDate = body.path("expectedArrivalAtPlaceOfDeliveryEndDate").asText("");
+      var invalidDates = new LinkedHashSet<String>();
+      if (!providedArrivalStartDate.isEmpty() && !providedArrivalEndDate.isEmpty()) {
```
HenrikHL, line 100: Does this catch the conditional check that if one is provided - then the other `MUST` be provided

nt-gt: This should probably be a stand-alone check. There is a `allOrNoneArePresent` that is specialized for this.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: "loadLocation and dischargeLocation as part of the Transport Plan " this needs to be there as well ?

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+  };
+  private static final Predicate<JsonNode> IS_ACTIVE_REEFER_SETTINGS_REQUIRED = (reqEquipNode) -> {
+    var norNode = reqEquipNode.path("isNonOperatingReefer");
+    if (norNode.isMissingNode() || !norNode.isBoolean()) {
+      // Only require the reefer if there is no equipment code or the equipment code is clearly a reefer.
+      // Otherwise, we give conflicting results in some scenarios.
+      return !HAS_ISO_EQUIPMENT_CODE.test(reqEquipNode) || IS_ISO_EQUIPMENT_CONTAINER_REEFER.test(reqEquipNode);
+    }
+    return !norNode.asBoolean(false);
+  };
```
HenrikHL, line 132: I am not sure this is correct.
It is correct that an `Active Reefer Settings` is required if `isNonOperatingReefer` is set to `false`. If `isNonOperatingReefer` is missing or if it is `true` then there cannot be a `Active Reefer Setting`. To me it looks like you allow a `Active Reefer Setting` if `isNonOperatingReefer` is missing

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        var commodities = requestedEquipments.get("commodities");
+        if (isShipperOwned) {
+          for (var commodity : commodities) {
+            var cargoGrossWeight = commodity.get("cargoGrossWeight");
+            var cargoGrossWeightUnit = commodity.get("cargoGrossWeightUnit");
+            var cargoGrossVolume = commodity.get("cargoGrossVolume");
+              var cargoGrossVolumeUnit = commodity.get("cargoGrossVolumeUnit");
+            if (cargoGrossWeight == null || cargoGrossWeight.isEmpty()) {
+              issues.add("the container weight be provided");
+            }
```
HenrikHL, line 160: Does this need to be explicitly - I would assume it would be caught by the schema validation rule...?

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: The `unique` condition requires the same element / key twice or more before it can trigger.  It does not sound like it was properly tested. Please try with a repeated version and report back.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        }
+      }
+      return issues;
+    }
+  );
+
+  private static final JsonContentCheck UNIVERSAL_SERVICE_REFERENCE = JsonAttribute.customValidator(
+    "Conditional Universal Service Reference",
+    (body) -> {
+      var expectedDepartureDate = body.path("universalExportVoyageReference");
```
HenrikHL, line 177: This is a copy paste error - the departure date is not fetched in the UniversalExportVoyage :-D

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+      }
+
+      return issues;
+    }
+  );
+
+  private static final JsonContentCheck REFERENCE_TYPE_VALIDATION = JsonAttribute.allIndividualMatchesMustBeValid(
+    "Validate reference type field",
+    (mav) -> {
+      mav.submitAllMatching("requestedEquipments.*.references.*.type");
```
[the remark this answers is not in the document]
preetamnpr: nope.. i just added it

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        .map(p -> p.path("cutOffDateTimeCode"))
+        .filter(JsonNode::isTextual)
+        .map(n -> n.asText(""))
+        .collect(Collectors.toSet());
+      if (receiptTypeAtOrigin.equals("CFS") && !cutOffDateTimeCodes.contains("LCL")) {
+        issues.add("cutOffDateTimeCode 'LCL' must be present when receiptTypeAtOrigin is CFS");
+      }
+      return issues;
+    }
+  );
```
HenrikHL, line 273: Please make sure that each cut-off time is only provided once (the `cutoffDateTimeCodes` in a booking must be unique)

HenrikHL: I can see you check this in line 284 :-)

preetamnpr: Line number 297
```
  private static final JsonContentCheck SHIPMENT_CUTOFF_TIMES_UNIQUE = JsonAttribute.allIndividualMatchesMustBeValid(
    "in 'shipmentCutOfftimes' cutOff date time code must be unique",
    ALL_SHIPMENT_CUTOFF_TIMES,
    JsonAttribute.unique("cutOffDateTimeCode")
  );
```

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    body -> {
+      var issues = new LinkedHashSet<String>();
+      var shipmentLocations = body.path("shipmentLocations");
+      var receiptTypeAtOrigin = body.path("receiptTypeAtOrigin").asText("");
+      var polNode =
+        StreamSupport.stream(shipmentLocations.spliterator(), false)
+          .filter(o ->  o.path("locationTypeCode").asText("").equals("POL")
+            || o.path("locationTypeCode").asText("").equals("PRE") )
+          .findFirst()
+          .orElse(null);
```
HenrikHL, line 306: This does not always fetch a POL. If both PRE and POL are present in the list - it is undefined which of them is returned in the polNode...?

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+      var podNode =
+        StreamSupport.stream(shipmentLocations.spliterator(), false)
+          .filter(o -> o.path("locationTypeCode").asText("").equals("POD")
+          || o.path("locationTypeCode").asText("").equals("PDE") )
+          .findFirst()
+          .orElse(null);
```
HenrikHL, line 312: Same problem

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: The shiplocations element can either be PRE or POL for PoL Node. I didnt get your question

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: can be omitted.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+    (mav) -> {
+      mav.submitAllMatching("documentParties.*.party.taxLegalReferences.*");
+    },
+    JsonAttribute.combineAndValidateAgainstDataset(BookingDataSets.LTR_CC_T_COMBINATIONS, "countryCode", "type")
+  );
+
+  private static final JsonContentCheck ISO_EQUIPMENT_CODE_VALIDATION = JsonAttribute.allIndividualMatchesMustBeValid(
+    "Validate ISO Equipment code",
+    (mav) -> {
+      mav.submitAllMatching("requestedEquipments.*.ISOEquipmentCode");
```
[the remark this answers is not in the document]
HenrikHL: `ISOEquipmentCodes` in `ConfirmedEquipments` needs to be checked as well

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
HenrikHL: My suggestion was to not call it `polNode` as it is sometimes a `PRE` - but never mind - the code works and your error message in line 321 is correct :-)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+  private static final JsonContentCheck REASON_PRESENCE = JsonAttribute.customValidator(
+    "Reason field must be present for the selected Booking Status",
+    (body) -> {
+      var bookingStatus = body.path("bookingStatus");
+      var amendedBookingStatus = body.path("amendedBookingStatus");
+      var issues = new LinkedHashSet<String>();
+      var status = amendedBookingStatus.isMissingNode() || amendedBookingStatus.isNull() ? bookingStatus : amendedBookingStatus;
+      if (REASON_STATES.contains(BookingState.fromWireName(status.asText()))) {
+        var reason = body.get("reason");
+        if (reason == null) {
```
HenrikHL, line 358: I am sure it was confirmed (somewhere on Teams) that it is not a requirement to add a `reason` - in some cases it is optional

preetamnpr: I am checking only the states where its must to have.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        }
+        if (body.get("charges") == null) {
+          issues.add("charges for confirmed booking is not present");
+        }
+        if (body.get("shipmentCutOffTimes") == null) {
+          issues.add("shipmentCutOffTimes for confirmed booking is not present");
+        }
+        if (body.get("advanceManifestFilings") == null) {
+          issues.add("advanceManifestFilings for confirmed booking is not present");
+        }
```
HenrikHL, line 392: This is not correct - it is only possible to have `advanceManifestFilings` in a confirmed Booking - but it is not a requirement

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        if (body.get("transportPlan") == null) {
+          issues.add("transportPlan for confirmed booking is not present");
+        }
+        if (body.get("carrierClauses") == null) {
+          issues.add("carrierClauses for confirmed booking is not present");
+        }
+        if (body.get("charges") == null) {
+          issues.add("charges for confirmed booking is not present");
+        }
```
HenrikHL, line 386: This is not correct - it is correct that `charges` can only be part of a Confirmed booking - but if there are no charges then it is ok to not provide the field (e.g. the field is not required)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+        if (body.get("termsAndConditions") == null) {
+          issues.add("termsAndConditions for confirmed booking is not present");
+        }
+        if (body.get("transportPlan") == null) {
+          issues.add("transportPlan for confirmed booking is not present");
+        }
+        if (body.get("carrierClauses") == null) {
+          issues.add("carrierClauses for confirmed booking is not present");
+        }
```
HenrikHL, line 383: This is not correct - it is correct that `clauses` can only be part of a Confirmed booking - but if there are no clauses then it is ok to not provide the field (e.g. the field is not required)

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: This check will only triggers for the confirmed booking states only.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -136,11 +144,20 @@ public static ActionCheck requestContentChecks(UUID matched, Supplier<CarrierSce
...
     )
   );
 
+  private static final JsonContentCheck ISO_EQUIPMENT_CODE_AND_NOR_CHECK = JsonAttribute.allIndividualMatchesMustBeValid(
+    "All requested Equipments where ISOEquipmentCode is reefer code must have 'isNonOperatingReefer' flag",
+    ALL_REQ_EQUIP,
+    JsonAttribute.ifMatchedThen(
+      IS_ACTIVE_REEFER_SETTINGS_REQUIRED,
+      JsonAttribute.path("isNonOperatingReefer", JsonAttribute.matchedMustBePresent())
+    )
```
nt-gt: If `IS_ACTIVE_REEFER_SETTINGS_REQUIRED` lives up to its name, then `isNonOperatingReefer` would always be false in this case. This strongly suggests that you want a different condition.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -136,11 +144,20 @@ public static ActionCheck requestContentChecks(UUID matched, Supplier<CarrierSce
     "All requested Equipments where 'isNonOperatingReefer' is 'false' must have 'activeReeferSettings'",
     ALL_REQ_EQUIP,
     JsonAttribute.ifMatchedThen(
-      IS_ACTIVE_REEFER_SETTINGS_REQUIRED,
+      IS_ISO_EQUIPMENT_CONTAINER_REEFER,
       JsonAttribute.path("activeReeferSettings", JsonAttribute.matchedMustBePresent())
     )
```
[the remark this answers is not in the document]
preetamnpr: You are right. Henrik wanted to add one more check on the Reefer. Its probably the undo changes or redo interchanged the condition.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -372,24 +407,9 @@ private static Consumer<MultiAttributeValidator> allDg(Consumer<MultiAttributeVa
...
         }
-        if (body.get("carrierClauses") == null) {
-          issues.add("carrierClauses for confirmed booking is not present");
-        }
-        if (body.get("charges") == null) {
-          issues.add("charges for confirmed booking is not present");
-        }
-        if (body.get("shipmentCutOffTimes") == null) {
-          issues.add("shipmentCutOffTimes for confirmed booking is not present");
-        }
```
nt-gt: @HenrikHL: I did not see you challenging this one being present on a confirmed booking, so I assume `shipmentCutOffTimes` is still required on a confirmed booking (and therefore this should not be removed).

core/src/main/java/org/dcsa/conformance/core/check/JsonAttribute.java
```diff
@@ -583,6 +583,9 @@ public static JsonContentMatchedValidation combineAndValidateAgainstDataset(
     return (nodeToValidate, contextPath) -> {
       var codeA = nodeToValidate.path(nameA).asText("");
       var codeB = nodeToValidate.path(nameB).asText("");
+      if (codeA.isEmpty() || codeB.isEmpty() ) {
+        return Set.of();
+      }
```
nt-gt, line 588: I am not ok with this change. The logic of this method was to validate that the fields combined matched against a data set. This changes exempts some cases if a field is absent/empty, but I have code using this method where this new exception should not apply. In fact, I am not sure this exception should ever apply.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
HenrikHL: Sorry - I misunderstood this check - have you made a check somewhere else that checks that a `reason` field should not be provided in a Confirmed Booking?

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,672 @@
...
+        if (body.hasNonNull("charges")) {
+          issues.add("charges must not be present".formatted(bookingStatus));
+        }
+        if (body.hasNonNull("shipmentCutOffTimes")) {
+          issues.add("shipmentCutOffTimes must not be present".formatted(bookingStatus));
+        }
+        if (body.hasNonNull("advanceManifestFilings")) {
+          issues.add("advanceManifestFilings must not be present".formatted(bookingStatus));
+        }
+      }
```
HenrikHL, line 404: `trasnportPlan` and `confirmedEquipments` should not be present in a non-confirmed booking

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -297,10 +297,17 @@ private static Consumer<MultiAttributeValidator> allDg(Consumer<MultiAttributeVa
...
+    && reqEquipNode.hasNonNull("manifestTypeCode");
+
+  private static final JsonRebaseableContentCheck AMF_CC_MTC_COMBINATION_VALIDATIONS = JsonAttribute.allIndividualMatchesMustBeValid(
+    "All utilizedTransportEquipments with a reefer ISO Equipment Code must have at least isNonOperatingReefer",
+    (mav) -> mav.submitAllMatching("advanceManifestFilings.*"),
+    JsonAttribute.ifMatchedThen(
+      HAS_NON_NULL_AMF_ITEM,
+      JsonAttribute.combineAndValidateAgainstDataset(BookingDataSets.AMF_CC_MTC_COMBINATIONS, "countryCode", "manifestTypeCode")
+    )
   );
```
nt-gt, line 311: To be honest, I do not understand why this logic is necessary.

The `submitAllMatching` on an array would should only match existing array elements and we should not have null elements.

Please show me the payload flagged as incorrect when you do not have this `HAS_NON_NULL_AMF_ITEM` guard in it.

booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
```diff
@@ -93,7 +93,8 @@ public void confirmBookingAmendment(String reference, String reason) {
     changeState(BOOKING_STATUS, CONFIRMED);
     changeState(AMENDED_BOOKING_STATUS, AMENDMENT_CONFIRMED);
     mutateBookingAndAmendment(this::ensureConfirmedBookingHasCarrierFields);
-    setReason(reason);
+    setReason(null);
+    removeRequestedChanges();
```
nt-gt, line 97: The `reason` parameter seems like it is now unused and can be removed.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
nt-gt: Still not done

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,645 @@
...
+      JsonAttribute.unique("partyFunction")
+    ));
+
+  private static final JsonContentCheck COMMODITIES_SUBREFERENCE_UNIQUE = JsonAttribute.allIndividualMatchesMustBeValid(
+    "Each Subreference in commodities must be unique",
+    (mav) -> {
+      mav.submitAllMatching("requestedEquipments.*.commodities");
+    },
+    JsonAttribute.unique("subReference")
+  );
```
nt-gt, line 260: Do we have a check that subreferences must be provided on the commodity when the booking is confirmed? (I cannot see one)

booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
```diff
@@ -93,7 +93,7 @@ public void confirmBookingAmendment(String reference, String reason) {
     changeState(BOOKING_STATUS, CONFIRMED);
     changeState(AMENDED_BOOKING_STATUS, AMENDMENT_CONFIRMED);
     mutateBookingAndAmendment(this::ensureConfirmedBookingHasCarrierFields);
```
nt-gt, line 95: The `reason` parameter is still unused in this method.

I assume the `setReason(...)` being removed is something you aligned with Henrik. I found it surprising, but if Henrik said so then it is fine.

preetamnpr: the reason is here an unused parameter. I will remove it . I checked with @HenrikHL (also IFS) that reason field is not present for booking states (the RECEIVED and CONFIRMED).

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: done

booking/src/main/java/org/dcsa/conformance/standards/booking/model/PersistableCarrierBooking.java
nt-gt: If the `reason` **must not** be present, then I suspect you want `setReason(null)` to ensure it is always cleared (it would also clear the reason field on confirming the booking, which probably makes sense either additionally in that case).

preetamnpr: it make sense to clear

HenrikHL:
My issues have been fixed (also the `isNonOperatingReefer` - so I am good now

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,716 @@
...
+      for (var rq: requestedEquipments) {
+        var isShipperOwned = rq.get("isShipperOwned").asBoolean(false);
+        var commodities = requestedEquipments.get("commodities");
+        if (isShipperOwned) {
+          for (var commodity : commodities) {;
+            var cargoGrossVolume = commodity.get("cargoGrossVolume");
+              var cargoGrossVolumeUnit = commodity.get("cargoGrossVolumeUnit");
+            if(cargoGrossVolume != null && (cargoGrossVolumeUnit  == null || cargoGrossVolumeUnit.isEmpty()) ) {
+              issues.add("the cargo gross volume unit must be provided");
+            }
```
nt-gt, line 180: Nit / For next PR: Spacing is awkward here.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,716 @@
...
+      var requestedEquipments = body.path("requestedEquipments");
+      for (var rq: requestedEquipments) {
+        var isShipperOwned = rq.get("isShipperOwned").asBoolean(false);
+        var commodities = requestedEquipments.get("commodities");
+        if (isShipperOwned) {
+          for (var commodity : commodities) {;
+            var cargoGrossVolume = commodity.get("cargoGrossVolume");
+              var cargoGrossVolumeUnit = commodity.get("cargoGrossVolumeUnit");
+            if(cargoGrossVolume != null && (cargoGrossVolumeUnit  == null || cargoGrossVolumeUnit.isEmpty()) ) {
+              issues.add("the cargo gross volume unit must be provided");
```
nt-gt, line 179: Bug (?): Why would we only check the `cargoGrossVolume` if it is shipper owned? This check does not make sense to me. The `cargoGrossVolume` vs. `cargoGrossVolumeUnit` looks like it should be pulled into a separate `JsonAttribute.presenceImpliesOtherField` (inside a `JsonAttribute.allIndividualMatchesMustBeValid`) without any `isShipperOwned` logic.

But then I am not sure what the remainder of this code is used for (tareWeight was my guess, but that is schema validated)...? Maybe it is just redundant.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,716 @@
...
+      var universalExportVoyageReference = body.path("universalExportVoyageReference");
+      var universalImportVoyageReference = body.path("universalImportVoyageReference");
+      var universalServiceReference = body.path("universalServiceReference");
+      var issues = new LinkedHashSet<String>();
+      if (JsonAttribute.isJsonNodePresent(universalExportVoyageReference) || JsonAttribute.isJsonNodePresent(universalImportVoyageReference) ) {
+        if (JsonAttribute.isJsonNodeAbsent(universalServiceReference) ){
+          issues.add("The universalServiceReference must be present as either universalExportVoyageReference or universalExportVoyageReference are present");
+        }
+      }
+      return issues;
```
nt-gt, line 200: Nit / For next PR:

Use `return Set.of("message ...");` / `return Set.of()`  instead of doing a `var issues = new LinkedHashSet<String>();` for cases like this where there is a 0-1 issue case (the pattern is fine if it is 0-N or a more complex check).

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: yes.. GrossVolume must be without shipperOwned. The containerTareWeight is for the ShipperOwned. Due to review process, I moved it out of this logic. I will check this one.

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,716 @@
...
+
+      if (!partyFunctions.contains("SCO")) {
+        if (!body.path("serviceContractReference").isTextual()) {
+          issues.add("The 'SCO' party is mandatory when 'serviceContractReference' is absent");
+        }
+        if (!body.path("contractQuotationReference").isTextual()) {
+          issues.add("The 'SCO' party is mandatory when 'contractQuotationReference' is absent");
+        }
+      }
+      Set<String> partyShippers  = new HashSet<>(Arrays.asList("OS", "DDR"));
```
nt-gt, line 545: Nit: The right hand side can be reduced to `Set.of("OS", "DDR");`

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: done

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: done

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
preetamnpr: done

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,716 @@
...
+
+  private static final JsonPointer CARRIER_BOOKING_REQUEST_REFERENCE = JsonPointer.compile("/carrierBookingRequestReference");
+  private static final JsonPointer CARRIER_BOOKING_REFERENCE = JsonPointer.compile("/carrierBookingReference");
+  private static final JsonPointer BOOKING_STATUS = JsonPointer.compile("/bookingStatus");
+  private static final JsonPointer[] BOOKING_UN_LOCATION_CODES = {
+    JsonPointer.compile("/invoicePayableAt/UNLocationCode"),
+    JsonPointer.compile("/placeOfBLIssue/UNLocationCode"),
+    JsonPointer.compile("/transportPlan/loadLocation/UNLocationCode"),
+    JsonPointer.compile("/transportPlan/dischargeLocation/UNLocationCode")
+  };
```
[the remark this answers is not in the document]
preetamnpr: done

booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -0,0 +1,684 @@
...
+      if (CONFIRMED_BOOKING_STATES.contains(BookingState.fromWireName(bookingStatus))) {
+        if (body.get("confirmedEquipments") == null) {
+          issues.add("confirmedEquipments for confirmed booking is not present");
+        }
+        if (body.get("transportPlan") == null) {
+          issues.add("transportPlan for confirmed booking is not present");
+        }
+        if (body.get("shipmentCutOffTimes") == null) {
+          issues.add("shipmentCutOffTimes for confirmed booking is not present");
+        }
```
nt-gt, line 408: I talked with Henrik, and the condition here is that must be non-empty.  I think the conditions can be written as `if (body.path("...").isEmpty())`

The presence check is insufficient as it enables / allows an empty list to be provided.

nt-gt (approved):

PR #82, opened 2024-04-03
Dt-863 abuse test cases implementation.

preetamnpr:


preetamnpr:
Merge to release branch until we have the final decision on releases of betas.

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,101 +78,131 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
   private BookingScenarioListBuilder thenAllPathsFrom(
       BookingState bookingState, BookingState originalBookingState) {
     return switch (bookingState) {
-      case CANCELLED, COMPLETED, DECLINED, REJECTED  -> then(shipper_GetBooking(bookingState));
+      case COMPLETED  -> then(
+        shipper_GetBooking(bookingState).thenEither(
+         uc7_shipper_submitBookingAmendment(true)
+            .then(shipper_GetBooking(bookingState))
+        )
```
nt-gt, line 85: Any reason why `cancel` is not being tried here as well?

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,101 +78,131 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
...
+      );
+      case DECLINED -> then(
+        shipper_GetBooking(bookingState).then(
+          uc7_shipper_submitBookingAmendment(true)
+            .then(shipper_GetBooking(bookingState))
+        )
```
nt-gt, line 91: Any reason why `cancel` is not being tried here as well?

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,101 +78,131 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
...
+      );
+      case CANCELLED  -> then(
+        shipper_GetBooking(bookingState).thenEither(
+          uc3_shipper_submitUpdatedBookingRequest(true)
+            .then(shipper_GetBooking(bookingState))
+        )
```
nt-gt, line 97: If `cancel` is not explicitly allowed on top of `cancel` then we should definitely have `cancel` here as well (for the return code check)

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,101 +78,131 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
   private BookingScenarioListBuilder thenAllPathsFrom(
       BookingState bookingState, BookingState originalBookingState) {
     return switch (bookingState) {
-      case CANCELLED, COMPLETED, DECLINED, REJECTED  -> then(shipper_GetBooking(bookingState));
+      case COMPLETED  -> then(
+        shipper_GetBooking(bookingState).thenEither(
+         uc7_shipper_submitBookingAmendment(true)
```
nt-gt, line 83: General remark: Please rename and use a different name for abuse test cases when they are used in the state transitions.

In the eBL, I named them `auc_X` (for `AbUse Case`).

this makes it easier to see at a distance that these transitions are invalid.


Note: I am fine with them being funneled into the same underling code (action classes) after that. The point is that I want to quickly be able to tell "valid" from "invalid" transition apart something as complex as this. A `true` / `false` will not cut it as `true` / `false` are non-descriptive (especially when the editor is not there to provide inlay hints, which it is not in GitHub among other)

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,101 +78,131 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
...
+            .then(shipper_GetBooking(bookingState))
```
nt-gt, line 84: Do we not check the state in `shipper_GetBooking`? For eBL the expected status is passed to the GET call.

Also, for an amendment call, you should also check that the amended booking is unchanged

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
preetamnpr: As per the story user performs UC7 when Booking status is COMPLETED and it hasn't been defined in the story.

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
preetamnpr: As per the story [DT-863](https://dcsa.atlassian.net/browse/DT-863) user performs UC7 when Booking status is DECLINED and it hasn't been defined in the story.

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
nt-gt: Challenge the story then to confirm whether this omission is intentional or an oversight

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
nt-gt: Challenge the story then to confirm whether this omission is intentional or an oversight

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -78,7 +80,24 @@ private BookingScenarioListBuilder thenAllPathsFrom(BookingState bookingState) {
...
+      case DECLINED, CANCELLED, REJECTED -> then(
+        shipper_GetBooking(bookingState).thenEither(
+          auc_shipper_sendInvalidBookingAction(UPDATE_BOOKING)
+            .then(shipper_GetBooking(bookingState)),
+          auc_shipper_sendInvalidBookingAction(SUBMIT_BOOKING_AMENDMENT)
+            .then(shipper_GetBooking(bookingState)),
+          auc_shipper_sendInvalidBookingAction(CANCEL_BOOKING)
+            .then(shipper_GetBooking(bookingState))
+        )
+      );
```
nt-gt, line 100: Missing a happy path flow (where there is no `auc` step)

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
```diff
@@ -109,16 +128,15 @@ private BookingScenarioListBuilder thenAllPathsFrom(
...
+                uc3_shipper_submitUpdatedBookingRequest()
+                    .thenAllPathsFrom(UPDATE_RECEIVED),
+                uc4_carrier_rejectBookingRequest().thenAllPathsFrom(REJECTED),
+                uc5_carrier_confirmBookingRequest().thenAllPathsFrom(CONFIRMED),
+                uc11_shipper_cancelBooking().thenAllPathsFrom(CANCELLED)));
       case START -> thenEither(
-        uc1_shipper_SubmitBookingRequest(ScenarioType.REGULAR).thenAllPathsFrom(RECEIVED),
-        uc1_shipper_SubmitBookingRequest(ScenarioType.REEFER).thenHappyPathFrom(RECEIVED),
-        uc1_shipper_SubmitBookingRequest(ScenarioType.DG).thenHappyPathFrom(RECEIVED));
+        uc1_shipper_SubmitBookingRequest().thenAllPathsFrom(RECEIVED));
```
nt-gt, line 139: I do not understand what made you remove the other scenario types?

preetamnpr: Good catch.. I was testing on local machine with only regular scenario hence I had to delete it.. will roll it back.

preetamnpr: I could recall now, those functions were redundant and not being used any more. Hence removed them.

booking/src/main/java/org/dcsa/conformance/standards/booking/BookingScenarioListBuilder.java
preetamnpr: done..

nt-gt (approved):

PR #96, opened 2024-05-31
DT-1124 changes

preetamnpr:


booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -235,6 +235,33 @@ private static Consumer<MultiAttributeValidator> allDg(Consumer<MultiAttributeVa
...
+        var partyContactDetails = childNode.path("partyContactDetails");
+        if (partyContactDetails.isArray()) {
+          StreamSupport.stream(partyContactDetails.spliterator(), false)
+            .forEach(element -> {
+              if (element.path("phone").isMissingNode() && element.path("email").isMissingNode()) {
+                issues.add("PartyContactDetails must have phone or an email id.");
+              }
+            });
+        }
+      }
```
nt-gt, line 261: This seems a bit convoluted. I think this can be simplified to use:

```
  private static final JsonContentCheck VALIDATE_DOCUMENT_PARTY =  JsonAttribute.allIndividualMatchesMustBeValid(
      ""Validate document party for address, identifyingCodes and partyContactDetails",
      (mav) -> mav.submitAllMatching("documentParties.*"),
      (childNode, contextPath) -> {
        var address = childNode.path("address");
        var identifyingCodes = childNode.path("identifyingCodes");
        if (address.isMissingNode() && identifyingCodes.isMissingNode()) {
          issues.add("address or identifyingCodes must have provided '%s'.".formatted(contextPath));
        }
        var partyContactDetails = childNode.path("partyContactDetails");
        if (partyContactDetails.isArray()) {
          StreamSupport.stream(partyContactDetails.spliterator(), false)
            .forEach(element -> {
              if (element.path("phone").isMissingNode() && element.path("email").isMissingNode()) {
                issues.add("PartyContactDetails must have phone or an email id at '%s[%d]'.".formatted(contextPath, -1));  // TODO; Replace -1 with the proper index
              }
            });
        }
        return issues;
      }

     ),
```

(Except all the bugs I introduced by writing it directly in GitHub without being able to test it).

Even if that is not an option, the `.fields()` seems weird because it feels like you are unpacking an object but `documentParties` is supposed to be an `ArrayNode` and not an `ObjectNode`. At the very least, I feel the code should use the constructs that match the expected node type.

preetamnpr: DocumentParties is an ObjectNode, not an ArrayNode. Hence I had to do come up with this logic otherwise I had the same logic in mind as you have written above.

  "documentParties": {
    "bookingAgent": {
      "partyName": "KN Bremerhaven",
.
.
.
.

https://app.swaggerhub.com/apis/dcsaorg/DCSA_BKG/2.0.0-Beta-3#/Booking/create-bookings

nt-gt: True, I forgot it was an object node.

Though, what about the `other` parties? It does not seem like this check properly covers that.

preetamnpr: aah. I didnt see. missed to look into. I will add this quickly.

nt-gt (approved):

PR #100, opened 2024-06-18
DT-1242 changes for regular dry container.

preetamnpr:


nt-gt (approved):

nt-gt (approved):

nt-gt:
FYI, I think George wanted this issue to also cover that the carrier can correctly echo/maintain all the attributes (Which we now have a check for that you can use)

preetamnpr:
@HenrikHL  reviewed the payload and here are his review comments: 

1. in Address object of BookingAgent - please remove name and please add stateRegion and UNLocationCode
2. in PartyContactDetails of BookingAgent - please add an email (it is an anyOf structure and allows for both phone and email)
3. in Booking Agent - please add multiple identifyingCodes and please add multiple taxLegalReferences
4. You also need to add more documentParties - preferably all optional ones (I know some are mutually exclusive - but shipper, serviceContractOwner, carrierBookingOffice, notifying parties, etc..)
5. in PartyContactDetails on Booking Root level - please add an email (it is an anyOf structure and allows for both phone and email)
6. for ShipmentLocations - please add one of each shipmentLocationType so we can validate that all are accepted. When doing so please use different location-interfaces so we don't always use UNLocationCode...
7. for RequestedEquipments - please add more objects where:
- also a isShipperOwned is true
- isNonOperatingReefer is TRUE so we test DRY cargo using Reefer equipment.
-  add references to RequestedEquipment
- Please add an OuterPackaging (I know it is only mandatory for DG - but it is optional for DRY). Only numberOfPackages should be provided and description

booking/src/main/java/org/dcsa/conformance/standards/booking/party/DynamicScenarioParameters.java
```diff
@@ -16,7 +16,9 @@ public record DynamicScenarioParameters(
     String carrierBookingRequestReference,
     String carrierBookingReference,
     BookingState bookingStatus,
-    BookingState amendedBookingStatus) {
+    BookingState amendedBookingStatus,
+    JsonNode booking
```
nt-gt, line 20: Are we not checking the amended booking request anywhere?

The `GET /booking/{ref}?amendedContent=true` (or whatever the query parameter is called)

preetamnpr: I am not sure what to check there, because the shipper modify the payload. The modified payload must be checked against the original Booking. If that's the case, then there would be some difference for sure w.r.t original booking and the checks fail. I tried this. Can you let me know what needs to done in order to check the amended Booking  request ?

nt-gt: > I am not sure what to check there, because the shipper modify the payload. The modified payload must be checked against the original Booking. If that's the case, then there would be some difference for sure w.r.t original booking and the checks fail. I tried this. Can you let me know what needs to done in order to check the amended Booking request ?

If the shipper requests an amendment to the booking and then uses `GET` to see their request again, the carrier should not have changed the amended variant between the request and the `GET` call. This implies tracking the current booking request and the amended booking requests separately and picking the right one for comparison based on the scenario step.

PR #105, opened 2024-07-12
DT-1316, DT-1328 and DT-1273 changes

preetamnpr:


booking/src/main/resources/standards/booking/messages/booking-api-2.0.0-regular.json
```diff
@@ -53,15 +53,35 @@
...
+          "name": "Port of Rotterdam",
+          "street": "Europoort Rotterdam",
+          "postCode": "3198 LK",
+          "city": "Rotterdam",
+          "countryCode": "NL"
+        },
+        "geoCoordinate": {
+          "latitude": "53.551086",
+          "longitude": "9.993682"
+        }
```
nt-gt, line 147: If we have this data, it should be a placeholder otherwise it will not match the UNLocationCode.

I recommend we do not have it for exactly that reason.

nt-gt: Note: It is fine for the "all attributes" message but not for the base API message used in all other tests (Given it was called `-regular` I assumed it was for the regular tests)

preetamnpr: I think it doesnt make sense to all addresses. I will remove them

nt-gt (approved):

PR #106, opened 2024-07-16
Remove references from SI that are no longer in the schema

nt-gt:


nt-gt:
As requested per mail

PR #109, opened 2024-07-26
STNG-129 Add first step of Manual testing to automated unit test

jkosternl:
It performs only the first scenario of the booking sandbox run. 

If you look in the build output of the `Checks` tab, you see 4 times `Tests run: ` and for each test, the success/failed/skipped numbers. You can see that my new test ran fine 😉


cdk/pom.xml
```diff
@@ -60,25 +59,5 @@
...
-          <version>${junit.version}</version>
-          <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.assertj</groupId>
-            <artifactId>assertj-core</artifactId>
-            <version>3.24.2</version>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
```
jkosternl, line 83: Totally not used in this module. Will add it later when needed.

preetamnpr (approved):

PR #115, opened 2024-08-09
DT-1216 202 changes

preetamnpr:


nt-gt (approved):

booking/src/main/java/org/dcsa/conformance/standards/booking/action/BookingAction.java
```diff
@@ -32,13 +33,27 @@ public BookingAction(
...
+    BookingAction previousAction,
+    String actionTitle,
+    Set<Integer> expectedStatus) {
+    super(sourcePartyName, targetPartyName, previousAction, actionTitle);
+    this.expectedStatus = expectedStatus;
+    this.dspReference =
+      previousAction == null
+        ? new OverwritingReference<>(null, new DynamicScenarioParameters(ScenarioType.REGULAR, null, null, null, null,null, null))
+        : new OverwritingReference<>(previousAction.dspReference, null);
+  }
```
jkosternl, line 55: I was just looking briefly at this PR and see 2 almost identical constructors. What is the reason of keeping 2? Can't we just only keep one, and make the callers just supplying a Set? (even if it is having only 1 value)

Btw, if you can use List as well, that's preferred above Set.

booking/src/main/java/org/dcsa/conformance/standards/booking/action/StateChangingBookingAction.java
```diff
@@ -3,12 +3,19 @@
...
+
 public abstract class StateChangingBookingAction extends BookingAction {
   public StateChangingBookingAction(String sourcePartyName, String targetPartyName, BookingAction previousAction,
                                     String actionTitle, int expectedStatus) {
     super(sourcePartyName, targetPartyName, previousAction, actionTitle, expectedStatus);
   }
 
+  public StateChangingBookingAction(String sourcePartyName, String targetPartyName, BookingAction previousAction,
+                                    String actionTitle, Set<Integer> expectedStatus) {
+    super(sourcePartyName, targetPartyName, previousAction, actionTitle, expectedStatus);
```
jkosternl, line 16: Same here. Please adjust the callers of the constructor to only use the changed Set<Integer> expectedStatus.
If you can use `List` instead, please also adopt.

jkosternl (changes requested):

booking/src/main/java/org/dcsa/conformance/standards/booking/action/BookingAction.java
jkosternl: Yes, I meant the `BookingAction` constructor indeed. There is only 1 or 2 usages of the single `expectedStatus` argument, so I don't see the need of keeping multiple constructor for every variance of constructing this object. The calling method can just easily be adapted to supplying a Set with a single item. That avoids code duplication.

booking/src/main/java/org/dcsa/conformance/standards/booking/action/StateChangingBookingAction.java
jkosternl: Yes, if most cases indeed are using one parameter on `expectedStatus`, than it makes sense. 
Using a Set makes sense if it we actively need to prevent duplicated values. I am fine with it, if it might happen.

jkosternl:
Preetam, if you are going to work on this PR, can you also please merge `dev` into your branch? That enables all the existing unit tests to happen and verify all existing sandbox scenarios are still working. Thanks!

jkosternl:
@preetamnpr, what should we do with this PR? You got some feedback; did you start a new PR because of that?

PR #116, opened 2024-08-19
STNG-131 Re-enable tests of sandboxIds

jkosternl:
and: Remove Carrier1Prompt and Platform1Prompt tests. Disable confusing API tests, which throw useless timeouts.

### DIFF
--- a/spring-boot/src/main/java/org/dcsa/conformance/springboot/ConformanceApplication.java
+++ b/spring-boot/src/main/java/org/dcsa/conformance/springboot/ConformanceApplication.java
@@ -202,9 +202,7 @@ public class ConformanceApplication {
 
   private void _addSimulatedLambdaDelay() {
     if (conformanceConfiguration.simulatedLambdaDelay > 0) {
-      log.info(
-          "Simulating lambda delay of %d milliseconds"
-              .formatted(conformanceConfiguration.simulatedLambdaDelay));
+      log.info("Simulating lambda delay of {} milliseconds", conformanceConfiguration.simulatedLambdaDelay);
       try {
         Thread.sleep(conformanceConfiguration.simulatedLambdaDelay);
       } catch (InterruptedException e) {
@@ -212,7 +210,7 @@ public class ConformanceApplication {
       }
       log.info("Done simulating lambda delay");
     } else {
-      log.info("No simulated lambda delay");
+      log.debug("No simulated lambda delay");
     }
   }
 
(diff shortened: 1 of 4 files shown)


spring-boot/src/main/java/org/dcsa/conformance/springboot/ConformanceApplication.java
```diff
@@ -202,17 +202,15 @@ public ConformanceApplication(ConformanceConfiguration conformanceConfiguration)
 
   private void _addSimulatedLambdaDelay() {
     if (conformanceConfiguration.simulatedLambdaDelay > 0) {
-      log.info(
-          "Simulating lambda delay of %d milliseconds"
-              .formatted(conformanceConfiguration.simulatedLambdaDelay));
+      log.info("Simulating lambda delay of {} milliseconds", conformanceConfiguration.simulatedLambdaDelay);
```
jkosternl, line 205: FYI: Using `{}` causes the logging framework to only use the given variable if the log level requires it. So, preventing formatting the String unnecessary.

spring-boot/src/test/java/org/dcsa/conformance/springboot/ConformanceApplicationTest.java
```diff
@@ -84,7 +80,7 @@ private void checkUntilScenariosAreReady(String sandboxId) throws InterruptedExc
         Thread.sleep(7_000L);
       }
     } while (!status.equals("{\"scenariosLeft\":0}"));
-    assertEquals("{\"scenariosLeft\":0}", status, "Scenario did not finish. Original start status: " + startStatus);
+    assertEquals("{\"scenariosLeft\":0}", status, "Scenario '" + sandboxId + "' did not finish properly! Original start status: " + startStatus);
```
jkosternl, line 83: It was confusing to see which `sandboxId` really failed. Adding this should make it very clear, I hope.

spring-boot/src/test/java/org/dcsa/conformance/springboot/ConformanceBasicAPITest.java
```diff
@@ -88,28 +82,6 @@ void shouldReturnPrintableReport() throws Exception {
       .andExpect(content().string(containsString("SupplyCSP [REGULAR]")));
   }
 
-  @Test
```
jkosternl, line 91: Already removed these, since the entry points are going to removed as well, as we discussed.

PR #120, opened 2024-08-26
Code cleanups

jkosternl:
No functional changes; see commit messages

### DIFF
--- a/core/src/main/java/org/dcsa/conformance/core/AbstractComponentFactory.java
+++ b/core/src/main/java/org/dcsa/conformance/core/AbstractComponentFactory.java
@@ -211,7 +211,7 @@ public abstract class AbstractComponentFactory {
     return "%s-%s-%s"
         .formatted(
             standardName.replace(" ", ""),
-            standardVersion.replaceAll("\\.", "").replace("-", ""),
+            standardVersion.replace("\\.", "").replace("-", ""),
             scenarioSuite.replace(" ", "-"))
         .toLowerCase();
   }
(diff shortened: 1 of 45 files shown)


core/src/main/java/org/dcsa/conformance/core/AbstractComponentFactory.java
```diff
@@ -210,9 +210,9 @@ protected static String _sandboxIdPrefix(
       String standardName, String standardVersion, String scenarioSuite) {
     return "%s-%s-%s"
         .formatted(
-            standardName.replaceAll(" ", ""),
-            standardVersion.replaceAll("\\.", "").replaceAll("-", ""),
-            scenarioSuite.replaceAll(" ", "-"))
+            standardName.replace(" ", ""),
+            standardVersion.replaceAll("\\.", "").replace("-", ""),
+            scenarioSuite.replace(" ", "-"))
```
jkosternl, line 215: `replace` also works on all occurrences. See e.g. https://medium.com/@KisinaDenis/what-is-the-difference-between-replace-vs-replaceall-method-in-java-string-class-6ca0e757ee5b

preetamnpr (approved):

core/src/main/java/org/dcsa/conformance/core/AbstractComponentFactory.java
jkosternl: Yes indeed. Will do that one too.

jkosternl: Directly addressed more `replaceAll` in the code base, without using regexp.

PR #122, opened 2024-08-29
STNG-145 Fix non-PR build: do not use PR parameters

jkosternl:
* Hard to test from a PR :-(

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -1,5 +1,10 @@
 name: Conformance-Gateway CI
 
+env:
+  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+  IN_PR: ${{ (github.event_name == 'pull_request' }} # SonarCloud
+  SONAR_PARAMS: -Dsonar.branch.name=$BRANCH_NAME
+
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/master.yml
```diff
@@ -42,13 +46,17 @@ jobs:
...
 
       - name: Run SonarCloud analysis
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }} # SonarCloud
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # SonarCloud
-          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         run: |
           cd Conformance-Gateway
-          mvn sonar:sonar -B -Dsonar.pullrequest.base=dev -Dsonar.pullrequest.key=${{github.event.pull_request.number}} -Dsonar.pullrequest.branch=$BRANCH_NAME
+          mvn sonar:sonar -B ${{env.SONAR_PARAMS}}
```
nt-gt, line 62: I think this would break and cause command line injection with branches that have spaces in them (the original version probably did too).

Maybe something like this

```shell
   if [ "${{github.event_name}}" == "pull_request" ]; then
        mvn sonar:sonar -B -Dsonar.pullrequest.base=dev -Dsonar.pullrequest.key="${{github.event.pull_request.number}}" -Dsonar.pullrequest.branch="$BRANCH_NAME"
   else
        mvn sonar:sonar -B -Dsonar.branch.name="$BRANCH_NAME"
   fi

```

And then remove `Set Sonar settings for PR build`?

nt-gt: I cannot produce any issues with the current setup and the proposed code might be safe. #124  and #125 were the attempts in case someone wants to give at a go.

PR #127, opened 2024-09-02
STNG-17 Improve GitHub CI Action

jkosternl:
The actual change was put in master (inside PR #128), but the new Action is not visible in GitHub. Trying to add the same commit from master into this PR. 

Also improving the existing master.yml file


.github/workflows/master.yml
```diff
@@ -24,28 +24,25 @@ jobs:
...
-          distribution: 'temurin'
-          java-version: 21
-          cache: 'maven'
-
       - name: Print commit sha
         shell: bash
         env:
           COMMIT_SHA: ${{ github.event.pull_request.head.sha }}
         run: |
           echo "sha: $COMMIT_SHA"
```
jkosternl, line 36: Just moved the 'Print commit sha' one place up, directly after the checkout.

PR #131, opened 2024-09-04
STNG-17 Make AWS deployments on dev branch run automatically

jkosternl:
* Disabled the manual input mode (for now)

### DIFF
--- a/.github/workflows/deploy-to-aws.yml
+++ b/.github/workflows/deploy-to-aws.yml
@@ -1,12 +1,6 @@
 name: Deploy to AWS
 
 on:
-#  workflow_dispatch:
-#    inputs:
-#      environment:
-#        description: 'The environment to deploy to'
-#        required: true
-#        default: 'dev'
   push:
     branches: ["dev"]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-to-aws.yml
```diff
@@ -1,18 +1,19 @@
...
-        required: true
-        default: 'dev'
-  #push:
-  #  branches: ["dev"]
+#  workflow_dispatch:
+#    inputs:
+#      environment:
+#        description: 'The environment to deploy to'
+#        required: true
+#        default: 'dev'
```
nt-gt, line 9: I would remove this rather than leaving it commented out if it is not longer useful.

nt-gt (approved):
LGTM with one minor remark

.github/workflows/deploy-to-aws.yml
jkosternl: Yes, makes sense. I'll wait until George approves this approach or not, before deleting it. Perhaps he'd like to keep the manual approach as well. Then I uncomment it.

PR #161, opened 2024-09-23
APCV-1367 and APCV-1268 - CS conformance issue fix

palatsangeetha:
-Fix for incorrect prompt and updated the GET actions title for pagination scenarios

### DIFF
--- a/commercial-schedules/src/main/java/org/dcsa/conformance/standards/cs/CsScenarioListBuilder.java
+++ b/commercial-schedules/src/main/java/org/dcsa/conformance/standards/cs/CsScenarioListBuilder.java
@@ -135,42 +135,42 @@ public class CsScenarioListBuilder extends ScenarioListBuilder<CsScenarioListBui
           "Limit and Pagination",
           noAction()
             .thenEither(
-              scenarioWithParametersPtpForPagination(getPtpRoutings(),PLACE_OF_RECEIPT,PLACE_OF_DELIVERY,LIMIT),
-              scenarioWithParametersPsForPagination(getPortSchedules(),UN_LOCATION_CODE,DATE,LIMIT),
-              scenarioWithParametersVsForPagination(getVesselSchedules(),VESSEL_IMO_NUMBER,LIMIT))))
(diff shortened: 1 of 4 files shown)


commercial-schedules/src/main/java/org/dcsa/conformance/standards/cs/CsScenarioListBuilder.java
```diff
@@ -208,10 +209,44 @@ private static CsScenarioListBuilder getPtpRoutings() {
...
+                getActionTitlePtp(hasNextAction, (CsAction) previousAction),
                 componentFactory.getMessageSchemaValidator("api", "pointToPointRoutings")));
   }
 
-  private static CsScenarioListBuilder getPortSchedules() {
+  private static String getActionTitlePs(boolean hasNextAction, CsAction previousAction) {
+    if (hasNextAction) {
+      if (previousAction instanceof CsGetPortSchedulesAction) {
+        return "GetPortSchedulesSecondPage";
+      } else {
```
jkosternl, line 220:
```suggestion
      }
```
Code behaves the same without the `else {` part, but looks a bit cleaner. Same for line 232 and 243.

jkosternl (approved):
Looking good, from a Java point of view. 
If you're not on a tight schedule on CS, adding an unit test to verify the second prompt is correct, helps securing this.

jkosternl:
Hi @palatsangeetha, if you're still coding in CS, can you also check this line in `CsAction` class? 
`if (!dsp.equals(updatedDsp)) dsp.set(updatedDsp);` because that `equals` method is comparing different types of objects, and therefore is always not equal. You can also improve that later, if you don't like it now.

palatsangeetha:
> Hi @palatsangeetha, if you're still coding in CS, can you also check this line in `CsAction` class? `if (!dsp.equals(updatedDsp)) dsp.set(updatedDsp);` because that `equals` method is comparing different types of objects, and therefore is always not equal. You can also improve that later, if you don't like it now.

@jkosternl Thanks for the comments on the CS PR, I will take care of it as part of a next change as changing that and then testing all flows might take some time.

douradofrazer:
@CodiumAI-Agent /improve

PR #168, opened 2024-09-26
STNG-190 Add Adoption API standard conformance testing

jkosternl:
No unit testing yet inside the new module, but all other test mechanisms are working on this new standard.


adoption/src/main/java/org/dcsa/conformance/standards/adoption/action/SupplyScenarioParametersAction.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  public boolean isInputRequired() {
+    return true;
+  }
+
+  @Override
+  public void handlePartyInput(JsonNode partyInput) {
+    super.handlePartyInput(partyInput);
+    suppliedScenarioParameters = SuppliedScenarioParameters.fromJson(partyInput.get("input"));
```
nt-gt, line 102: Possibly NPE. The `partyInput.get` can return  `null` and the callee does not seem to support it.

If you are sure it is never null, use `.required`. You can also use `.path` if "defaults" is fine for the callee.

As far as I can see, use of `JsonNode.get` instead of `.path` or `.required` is "almost always" a trap.

adoption/src/main/java/org/dcsa/conformance/standards/adoption/action/SupplyScenarioParametersAction.java
```diff
@@ -0,0 +1,104 @@
...
+        publisherPartyName,
+        null,
+        null,
+        "SupplyScenarioParameters(%s)"
+            .formatted(
+                Arrays.stream(filterParameters)
+                    .map(FilterParameter::getQueryParamName)
+                    .collect(Collectors.joining(", "))));
+    this.filterParameters =
+        Stream.of(filterParameters).collect(Collectors.toCollection(LinkedHashSet::new));
```
nt-gt, line 37: Could this be simplified to `new LinkedHashSet<>(Arrays.asList(filterParameters))`?

adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/AdoptionAdopter.java
```diff
@@ -0,0 +1,100 @@
...
+    syncCounterpartPut(PutAdoptionStatsAction.PUT_ADOPTION_STATS_URL, jsonResponseBody);
+
+    addOperatorLogEntry("Provided response body, contents of file: %s".formatted(fileName));
+  }
+
+  @Override
+  public ConformanceResponse handleRequest(ConformanceRequest request) {
+    JsonNode jsonResponseBody =
+        JsonToolkit.templateFileToJsonNode(
+            AdoptionStandard.ADOPTION_STATS_EXAMPLE.formatted(apiVersion), Map.ofEntries());
```
nt-gt, line 93: The example has a hard-coded date and interval, so this response does not respect the request parameters. I would have expected the response to at least be aligned with the interval type and range of the input.

adoption/src/main/resources/standards/adoption/messages/adoption-1.0.0-example.json
```diff
@@ -0,0 +1,60 @@
+{
+  "lastUpdatedDateTime": "2024-06-18T20:44:37+01:00",
+  "interval": "day",
+  "startDate": "2024-06-16",
+  "endDate": "2024-06-22",
```
nt-gt, line 5: Is this correct? It looks like the original example that you questioned when we talked about how to do this scenario.

jkosternl: You're right, it is not correct indeed. George and me talked about the standard this afternoon and he is going to update the spec to (or ask Henrik to do so).

adoption/src/main/java/org/dcsa/conformance/standards/adoption/action/SupplyScenarioParametersAction.java
jkosternl: Yes indeed. This class I didn't give much attention actually.

adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/AdoptionAdopter.java
jkosternl: Yes, you're right! Well spotted. 

This standard is a little different than others, in a way that the role Adopter and role DCSA are developed. But for the role DCSA, we don't really care much, because we're not going to test our own implementation with it (George told me). So, that's why I didn't spend time in that one.

adoption/src/main/java/org/dcsa/conformance/standards/adoption/action/SupplyScenarioParametersAction.java
jkosternl: @gj0dcsa, just to make sure I got you right: I just checked, and if I supply `"interval": "asdf"` that also still makes it continue and send it to the user. Is it the right place to add validation of this input data, in this method? If you have an example otherwise, please point it out.

douradofrazer:
@CodiumAI-Agent /improve

douradofrazer:
@CodiumAI-Agent /review
--pr_reviewer.inline_code_comments=true
--pr_reviewer.require_score_review=true

PR #173, opened 2024-09-27
DT-1540 Move to http code 202

preetamnpr:


booking/src/main/java/org/dcsa/conformance/standards/booking/checks/BookingChecks.java
```diff
@@ -659,10 +659,18 @@ private static void generateScenarioRelatedChecks(List<JsonContentCheck> checks,
...
+      )
+    ),
+    JsonAttribute.allIndividualMatchesMustBeValid(
+      "The 'cargoGrossWeight' implies 'cargoGrossWeightUnit'",
+      mav -> mav.submitAllMatching("requestedEquipments.*.commodities.*.cargoGrossWeight"),
+      JsonAttribute.presenceImpliesOtherField(
+        "value",
+        "unit"
       )
     ),
```
nt-gt: I do not think this is necessary? I thought `value` and `unit` were both mandatory in this kind of object.

nt-gt: Nvm, fixed in a later commit

nt-gt (approved):

PR #177, opened 2024-10-01
APCV-1377-Updated the CS schema with the latest one

palatsangeetha:
Updated CS conformance with the latest API schema


jkosternl (approved):
(assuming that the build check is green, after a while)

PR #178, opened 2024-10-02
DT-1657: Improve validation related to number of copies/originals

nt-gt:
The issue is about copies because I swapped the copies and originals issue (the fix for other issue was already reviewed and merged despite being for the opposite attributes). Keeping the confusing total, this commits to the issue swapped problem.


palatsangeetha (approved):
The changes looks good.

PR #182, opened 2024-10-07
STNG-165 Run one scenario in AWS dev fully end-to-end

jkosternl:
Created test `AWSEnvironmentTest` which runs Ebl, Conformance TD-only, Carrier fully end-to-end. 
Created the SeleniumTestBase for code reuse. It does not start Spring Boot itself. 

Sorry for the refactoring and changes in the same PR.. which makes it harder to see the changes.

### DIFF
--- a/.github/workflows/deploy-to-aws.yml
+++ b/.github/workflows/deploy-to-aws.yml
@@ -54,4 +54,4 @@ jobs:
           TEST_LOGIN_PASSWORD: ${{ secrets.AWS_DEV_LOGIN_PASSWORD }}
           TEST_LOGIN_URL: ${{ secrets.AWS_DEV_WEB_URL }}
         run: |
-          mvn -B -U -Djunit.includedTags=Integration -Djunit.excludedTags=WebUI test
+          mvn -B -Djunit.includedTags=Integration -Djunit.excludedTags=WebUI test -X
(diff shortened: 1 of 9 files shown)


.github/workflows/deploy-to-aws.yml
```diff
@@ -16,9 +16,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          ref: ${{ env.AWS_ENVIRONMENT }}
-          # Enforces only AWS env name matching branches are checkout and deployed.
```
jkosternl, line 21: This was a pain.. I was making experimental changes in the `pom.xml` file for the new situation, but they were not applied, because I changed the checkout here to force it to `dev` HEAD instead. 
The enforcing of the right environment is also already guaranteed by line 5 and line 10 of this file.

pom.xml
```diff
@@ -150,10 +148,36 @@
...
+			</properties>
+		</profile>
+		<profile>
+			<id>integration</id>
+			<properties>
+				<junit.includedTags>Integration</junit.includedTags>
+				<junit.excludedTags>WebUI</junit.excludedTags>
+			</properties>
+		</profile>
+	</profiles>
```
jkosternl, line 179: I introduced profiles to make it more clear and easy to use and select which profile you'd like to use. By default, the first profile is used.

palatsangeetha (approved):

PR #183, opened 2024-10-08
DT-1821-Updated the issuance schema and the request json to the latest

palatsangeetha:


ebl-issuance/src/main/resources/standards/eblissuance/messages/eblissuance-v3.0.0-request.json
```diff
@@ -102,65 +181,26 @@
...
   },
   "issueTo": {
-    "sendToPlatform": "SEND_TO_PLATFORM_PLACEHOLDER",
     "partyName": "ISSUE_TO_LEGAL_NAME_PLACEHOLDER",
-    "partyCodes": [
+    "sendToPlatform": "SEND_TO_PLATFORM_PLACEHOLDER",
+    "identifyingCodes": [
       {
-        "partyCode": "ISSUE_TO_PARTY_CODE_PLACEHOLDER",
         "codeListProvider": "EPUI",
```
nt-gt, line 202: @palatsangeetha: The `EPUI` code is no longer valid for this field. I am a bit surprised that the CTK is giving green for this. I suspect one of the eBL conditional validations are not up to date (the same code is used for eBL, ISS, and PINT, so the bug is in all the standards).


@gj0dcsa: FYI in case you were planning to deploy it and someone test.

palatsangeetha: > @palatsangeetha: The `EPUI` code is no longer valid for this field. I am a bit surprised that the CTK is giving green for this. I suspect one of the eBL conditional validations are not up to date (the same code is used for eBL, ISS, and PINT, so the bug is in all the standards).
> 
> @gj0dcsa: FYI in case you were planning to deploy it and someone test.

Thanks @nt-gt for pointing this out. I will update the request with example values from the latest API spec.

PR #184, opened 2024-10-08
DT-1821-Updated the request json examples from the latest API spec

palatsangeetha:


ebl-issuance/src/main/resources/standards/eblissuance/messages/eblissuance-v3.0.0-request.json
```diff
@@ -87,70 +87,86 @@
...
         },
         "identifyingCodes": [
           {
-            "codeListProvider": "SMDG",
-            "codeListName": "LCL",
-            "partyCode": "HMM"
+            "codeListProvider": "W3C",
+            "partyCode": "MSK",
+            "codeListName": "DID"
           }
```
nt-gt, line 148: The party code is for MSK (and should probably not use `W3C` + `DID`. It looks like an `SMDG` code instead). However, the `partyName` is for Hyundai (HMM).

I assume the original example is wrong. Probably the issuing party should be made a placeholder that the issuer can provide.

ebl-issuance/src/main/resources/standards/eblissuance/messages/eblissuance-v3.0.0-request.json
```diff
@@ -87,70 +87,86 @@
...
       },
       "consignee": {
         "partyName": "CONSIGNEE_LEGAL_NAME_PLACEHOLDER",
         "identifyingCodes": [
           {
-            "codeListProvider": "EPUI",
+            "codeListProvider": "W3C",
             "partyCode": "CONSIGNEE_PARTY_CODE_PLACEHOLDER",
             "codeListName": "CONSIGNEE_CODE_LIST_NAME_PLACEHOLDER"
           }
```
nt-gt, line 123: I think these need to be turned into a placeholder objects. It is exceedingly likely that the parties do **not** have a fixed `codeListProvider` for this party.

I feel this basically applies to all parties with placeholders in them.

nt-gt:
The more I think about this, the more I feel this ends up overlapping with DT-1832 (assigned to me). I am not sure how best to go about this (how best to resolve the overlap). Maybe we just cut my comments out and leave them for me to deal with as a part of DT-1832 and then move forward with this as-is.

palatsangeetha:
> The more I think about this, the more I feel this ends up overlapping with DT-1832 (assigned to me). I am not sure how best to go about this (how best to resolve the overlap). Maybe we just cut my comments out and leave them for me to deal with as a part of DT-1832 and then move forward with this as-is.

Thanks Niels. I will go ahead and merge this PR then.

PR #194, opened 2024-10-14
DT-1844 - EBL: bring inputRequired and JsonForHumanReadablePrompt inline in UC6

jkosternl:
I think this is the fix Niels.


nt-gt:
For me, the inversion of the boolean checks seem off to me.

The only real change I see is that we are apparently required to manually return null from `getJsonForHumanReadablePrompt` even when `isInputRequired` returns `false`. I guess that is the bit that we were not aware of.

nt-gt:
I will take it from here (I only needed to know what we were missing and now I do)

jkosternl:
@nt-gt Yes indeed, because the front-end behaves differently on the return values of those Java classes.

PR #196, opened 2024-10-15
DT-1844: Align prompt with isInputRequired

nt-gt:


preetamnpr (approved):

nt-gt:
@jkosternl This should be one blocker of your list for #189

PR #207, opened 2024-10-22
Apcv 913

preetamnpr:


tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/DynamicScenarioParameters.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import lombok.NonNull;
+import lombok.With;
+
+import static org.dcsa.conformance.core.toolkit.JsonToolkit.OBJECT_MAPPER;
+
+@With
+public record DynamicScenarioParameters(
```
jkosternl, line 11: Hi @preetamnpr, in the commit of yesterday afternoon, with the message "STNG-184 Improve all DSP and related classes (#198)" I've improved the DSP code in our whole code base. Can you please checkout those examples? You can do the same for this TNT DSP impl. Let me know if I can help.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
```diff
@@ -112,13 +127,153 @@ private void supplyScenarioParameters(JsonNode actionPrompt) {
...
+    attributeMappings.put("eventCreatedDateTime:gt", List.of(new AttributeMapping(JsonPointer.compile("/eventCreatedDateTime"), (node, value) -> !node.isMissingNode()
+      && (OffsetDateTime.parse(node.asText()).isAfter(OffsetDateTime.parse(value))), Set.of())));
+    attributeMappings.put("eventCreatedDateTime:lt", List.of(new AttributeMapping(JsonPointer.compile("/eventCreatedDateTime"), (node, value) -> !node.isMissingNode()
+      &&  (OffsetDateTime.parse(node.asText()).isBefore(OffsetDateTime.parse(value))), Set.of())));
+    attributeMappings.put("eventCreatedDateTime:lte", List.of(new AttributeMapping(JsonPointer.compile("/eventCreatedDateTime"), (node, value) -> !node.isMissingNode()
+      && (OffsetDateTime.parse(node.asText()).isBefore(OffsetDateTime.parse(value)) || OffsetDateTime.parse(node.asText()).isEqual(OffsetDateTime.parse(value))), Set.of())));
+    attributeMappings.put("eventCreatedDateTime:eq", List.of(new AttributeMapping(JsonPointer.compile("/eventCreatedDateTime"), (node, value) -> !node.isMissingNode()
+      && OffsetDateTime.parse(node.asText()).isEqual(OffsetDateTime.parse(value)), Set.of())));*/
```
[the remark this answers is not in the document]
preetamnpr: Hi George, I have pushed the final changes and they are refactored.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/DynamicScenarioParameters.java
jkosternl: Yesterday afternoon, the PR was merged. Please compare this class with any other DSP class. You can reduce this class now, to only a few lines. Please let me know if I can help @preetamnpr

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/QueryParameterSpecificRule.java
```diff
@@ -0,0 +1,68 @@
...
+public class QueryParameterSpecificRule implements QueryParamRule{
+
+  private final Map<String, Set<String>> queryParamMap = Map.of(
+      "SHIPMENT", Set.of("shipmentEventTypeCode", "documentTypeCode", "carrierBookingReference",
+      "transportDocumentID", "transportDocumentReference", "equipmentReference", "eventCreatedDateTime"),
+      "TRANSPORT", Set.of("transportDocumentReference", "transportEventTypeCode", "transportCallID", "vesselIMONumber",
+      "exportVoyageNumber", "carrierServiceCode", "UNLocationCode", "equipmentReference", "eventCreatedDateTime"),
+      "EQUIPMENT", Set.of("carrierBookingReference", "transportDocumentReference", "transportCallID",
+      "vesselIMONumber", "exportVoyageNumber", "carrierServiceCode", "UNLocationCode", "equipmentEventTypeCode", "equipmentReference", "eventCreatedDateTime")
+    );
```
jkosternl, line 23: These 2 final class fields (queryParamMap & excludedQueryParams) can be made static, so they are not build up upon every initiation. They don't change, so it is fine. 
Can you also format this class?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/SortCriteria.java
```diff
@@ -0,0 +1,18 @@
...
+
+import lombok.Getter;
+
+import java.util.List;
+
+@Getter
+public class SortCriteria {
+  private final String field;
+  private final List<AttributeMapping> attributeMappings;
+  private final SortDirection direction;
```
jkosternl, line 11: I think you can use the Java Record keyword here as well. You don't need the constructor then. And records are automatically immutable. See e.g. https://www.baeldung.com/java-record-keyword

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
```diff
@@ -112,13 +130,167 @@ private void supplyScenarioParameters(JsonNode actionPrompt) {
   public ConformanceResponse handleRequest(ConformanceRequest request) {
     log.info("TntPublisher.handleRequest(%s)".formatted(request));
 
-    JsonNode jsonResponseBody =
-        JsonToolkit.templateFileToJsonNode(
-            "/standards/tnt/messages/tnt-220-response.json", Map.ofEntries());
+    Map<String, List<AttributeMapping>> attributeMappings = AttributeMapping.initializeAttributeMappings();
+
+    ObjectMapper mapper = new ObjectMapper();
```
jkosternl, line 135: Please use the existing `OBJECT_MAPPER` object, which you can import. Creating the ObjectMapper every time is taking resources.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
```diff
@@ -112,13 +130,167 @@ private void supplyScenarioParameters(JsonNode actionPrompt) {
...
+                  || results.stream().anyMatch(result -> mapping.getValues().contains(result.asText())))) {
+                  eventIds.add(eventId);
+                  filteredArray.add(node);
+                }
+              }
+            }
+          }
+        }
+      }
+    }
```
jkosternl, line 172: This whole code block above (line 147 until 172) is quite complex and hard to understand what is going on. Can yo u e.g. extract this as a method? That already helps.

By the way, is this logic quite generic? To filter out parameters from a `ConformanceRequest`? Should we have/use (add?) something in core module, to do this?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
```diff
@@ -112,13 +130,167 @@ private void supplyScenarioParameters(JsonNode actionPrompt) {
...
+
+  private ConformanceResponse return404(ConformanceRequest request, String message) {
+    return request.createResponse(
+      404,
+      Map.of(API_VERSION, List.of(apiVersion)),
+      new ConformanceMessageBody(createErrorResponse(request.method(), request.url(), "Not Found", message)));
+  }
+
+  private ObjectNode createErrorResponse(String httpMethod, String requestUri, String reason, String message) {
+    ObjectMapper mapper = new ObjectMapper();
```
jkosternl, line 245: Please remove this new mapper, and use `OBJECT_MAPPER`.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/SortDirection.java
```diff
@@ -0,0 +1,13 @@
+package org.dcsa.conformance.standards.tnt.party;
+
+public enum SortDirection {
+  ASCENDING("ASC"),
+  DESCENDING("DESC");
```
jkosternl, line 5: Is it possible to just use `ASC, DESC` instead of different name & value? Or are both really needed for the end user or in the standard? Just curious.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/AttributeMapping.java
```diff
@@ -0,0 +1,160 @@
...
+  private static final String DOCUMENT_TYPE_CODE = "documentTypeCode";
+  private static final String EVENT_CREATED_DATE_TIME = "eventCreatedDateTime";
+
+  public AttributeMapping(String jsonPath, BiPredicate<JsonNode, String> condition, Set<String> values) {
+    this.jsonPath = jsonPath;
+    this.condition = condition;
+    this.values = values;
+  }
+
+  public static Map<String, List<AttributeMapping>> initializeAttributeMappings() {
```
jkosternl, line 24: This method actually generates the attribute mappings, not only the init, but it actually returns it. That should be reflected in the name, such as getAttributeMappings?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
```diff
@@ -0,0 +1,79 @@
...
+        continue;
+      }
+      JsonSchemaValidator eventSchemaValidator = eventSchemaValidators.get(eventType);
+      for (String validationError : eventSchemaValidator.validate(eventNode)) {
+        validationErrors.add("Event #%d: %s".formatted(eventIndex, validationError));
+      }
+    }
+    return validationErrors;
+  }
+}
```
jkosternl, line 79: Can you please format this class?
Is it also possible to stream over `eventNodes` around line 59, instead of a for-loop?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
```diff
@@ -0,0 +1,79 @@
...
+      .formatted(HttpMessageType.RESPONSE.name().toLowerCase()), isRelevantForRoleName,
+      matchedExchangeUuid, httpMessageType);
+    this.eventSchemaValidators = eventSchemaValidators;
+  }
+
+  private boolean isEventNode(JsonNode jsonNode) {
+    return jsonNode.isObject() && !jsonNode.path("eventType").isMissingNode();
+  }
+
+  private ArrayList<JsonNode> _findEventNodes(
```
jkosternl, line 36: Please use the default naming convention of Java, so no `_` before method names, just camel Casing. 
See https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntChecks.java
```diff
@@ -0,0 +1,389 @@
...
+          .map(node -> node.path("eventType").asText())
+          .filter(eventType -> !VALID_EVENT_TYPES.contains(eventType))
+          .collect(Collectors.toSet());
+
+        if (!invalidEventTypes.isEmpty()) {
+          return invalidEventTypes.stream()
+            .map(eventType -> "Invalid eventType: " + eventType)
+            .collect(Collectors.toSet());
+        }
+        return Set.of();
```
jkosternl, line 83: What happens if you remove the `if (!invalidEventTypes.isEmpty()) {` line and just stream over the empty Set? It also returns an empty Set, I would assume, which is fine right? If that works, just remove line 83. Same for the other checks below and above.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/DynamicScenarioParameters.java
preetamnpr: Sure. I am yet to do the pagination implementation and I would take the latest changes for DSP.

jkosternl (changes requested):
Good work Preetam. I've added mostly small remarks, which are just cleaning some code and making it a bit easier to read and understand. Please let me know if you have questions. Also, click at the "refresh" icon in the main PR screen, after you've committed the changes please. Then I know, I need to recheck.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class TntSchemaConformanceCheck extends ActionCheck {
+
+  private final Map<TntEventType, JsonSchemaValidator> eventSchemaValidators;
+  public TntSchemaConformanceCheck(
```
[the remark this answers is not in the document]
preetamnpr: @gj0dcsa  I have created the TntChecks.java where all the response checks are performed. I didnt remove this rather i moved your logic to the different file called TntSchemaConformanceCheck.java. If you agree, I will move this logic to TntChecks.java.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/DynamicScenarioParameters.java
jkosternl: Great, thanks! 👍🏻

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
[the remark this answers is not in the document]
preetamnpr: > There's also the problem that if this finds no events, it returns success, which it shouldn't, so please return a validation error stating that no events (of the expected type) were found in the message.

regarding on your above comment in the (
https://github.com/dcsaorg/Conformance-Gateway/pull/207#discussion_r1814439434
). If there are no events sent by carrier for the Get Events then its a valid response.  I had to confirm with the KP on this during implementation.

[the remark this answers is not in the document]
preetamnpr: > Then please let's fix my original design as follows: if this custom lenient validation returns any errors, we return those. If it doesn't return any errors though, then let's perform the usual strict schema validation that we do for the other standards, and reports the results of that.

on this as well, I have created the TntChecks.java which does all query parameters check and attribute check in Json Response (tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntChecks.java). Please take a look into this.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/QueryParameterSpecificRule.java
preetamnpr: this is done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/SortCriteria.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/TntPublisher.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntSchemaConformanceCheck.java
preetamnpr: done

tnt/src/main/java/org/dcsa/conformance/standards/tnt/checks/TntChecks.java
preetamnpr: the error message needs to be provided back to the subChecks, hence i need to have it.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/TntScenarioListBuilder.java
```diff
@@ -114,7 +130,21 @@ private static TntScenarioListBuilder getEvents() {
             new TntGetEventsAction(
                 subscriberPartyName,
                 publisherPartyName,
-                previousAction,
+              (TntAction) previousAction,
                 componentFactory.getEventSchemaValidators()));
   }
+
+  private static TntScenarioListBuilder getEventsBadRequest() {
+    TntComponentFactory componentFactory = threadLocalComponentFactory.get();
```
palatsangeetha, line 138: Just trying to understand, why do we need the bad request as a separate action?

palatsangeetha: Can't we accomplish this by passing the GetEvents action with isBadRequest? Just checking as it appeared as a scenario than an action to me. It's totally fine with me if the workflow demands it to be like this. You can ignore this comment.

preetamnpr: I even thought about it but too many booleans travelling across the classes makes code bit unreadable. Hence I used this option also the same in booking as well.

palatsangeetha (approved):

preetamnpr:
> When I run T&T in auto mode in your branch, I get a single group "All scenarios". Please group the scenario into 3 distinct groups, one for each event type.

 @gj0dcsa  Sure,  this one is on my list once the reports were published. I did these changes and I will push the changes now.

preetamnpr:
> > When I run T&T in auto mode in your branch, I get a single group "All scenarios". Please group the scenario into 3 distinct groups, one for each event type.
> 
> @gj0dcsa Sure, this one is on my list once the reports were published. I did these changes and I will push the changes now.

Here are the snapshot of the report.
image

PR #216, opened 2024-11-04
SD-335 Add signature validation in eblissuance

palatsangeetha:


jkosternl:
@palatsangeetha, it is looking good. But did you also check the Selenium test for Issuance? I guess that one fails now. We probably need to make the same adjustment as what I did today for PINT, in the `SeleniumTestBase` class.

palatsangeetha:
> @palatsangeetha, it is looking good. But did you also check the Selenium test for Issuance? I guess that one fails now. We probably need to make the same adjustment as what I did today for PINT, in the `SeleniumTestBase` class.

Thanks @jkosternl for the input, have added the same to 'seleniumTestBase` class and the tests are running fine.

jkosternl (approved):

jkosternl (approved):

PR #218, opened 2024-11-05
SD-1575 Convert schemas from JSON to YAML

jkosternl:
Notes:

- Converted all JSON schemas, except for JIT which will be addressed later anyways. 
- For CS and OVS: I needed to add an array of objects, to get it working. These were the only manual changes. Those standards are published, so the original version can not be updated with that. 
- Review note: you don't need to check the YAML files itself, while reviewing. Please check the other changes. 
- Upgraded the library `json-schema-validator` to the latest version, even including my own changes 😄


ovs/src/main/java/org/dcsa/conformance/standards/ovs/OvsComponentFactory.java
```diff
@@ -98,11 +98,7 @@ public Set<String> getReportRoleNames(
   }
 
   public JsonSchemaValidator getMessageSchemaValidator(String apiProviderRole, boolean forRequest) {
-    String schemaFilePath =
-        "/standards/ovs/schemas/ovs-%s-%s.json"
-            .formatted(
-                standardVersion.toLowerCase().replaceAll("[.-]", ""),
-                apiProviderRole.toLowerCase());
+    String schemaFilePath = "/standards/ovs/schemas/OVS_v%s.yaml".formatted(standardVersion);
```
preetamnpr, line 101: indentation ?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/QueryParameterSpecificRule.java
```diff
@@ -1,18 +1,17 @@
 package org.dcsa.conformance.standards.tnt.party;
 
+import static org.dcsa.conformance.standards.tnt.checks.TntDataSets.VALID_DOCUMENT_TYPE_CODES;
```
preetamnpr, line 3: the guidelines  suggest pacakge imports first and then the static imports.

preetamnpr (changes requested):

ebl-surrender/src/main/java/org/dcsa/conformance/standards/eblsurrender/party/EblSurrenderCarrier.java
```diff
@@ -82,12 +82,26 @@ private void supplyScenarioParameters(JsonNode actionPrompt) {
     String tdr = UUID.randomUUID().toString().replace("-", "").substring(0, 20);
     eblStatesById.put(tdr, EblSurrenderState.AVAILABLE_FOR_SURRENDER);
     persistentMap.save("response", actionPrompt.get("response"));
+
```
palatsangeetha, line 85: Why are these changes part of this PR?

jkosternl: Because I updated the standard with the latest schema from the OpenAPI repo. The contain updates, which were not included in the current JSON schema in dev.

core/src/test/resources/schemas/dcsaorg-DCSA_BKG-2.0.0-resolved.json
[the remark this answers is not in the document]
jkosternl: Yes, the path of this file shows it is only used in test code. The test class `JsonSchemaValidatorTest` uses it.

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/QueryParameterSpecificRule.java
jkosternl: I suggest that we keep following the way our IDE formats the imports nowadays and that we do not manually interfere. We're not reading it nor updating it manually, so it should not matter that much. 
If your IDE reorders the imports differently than mine, we should update the `.editorconfig` file, so it is kept consistent throughout all dev environments. I expected that it was already the case. 
Did you also install the `google-java-format` plugin AND did you enable it in our project @preetamnpr ?

palatsangeetha (approved):

tnt/src/main/java/org/dcsa/conformance/standards/tnt/party/QueryParameterSpecificRule.java
jkosternl: Since I don't see you online and we have the sprint planning soon, I am proceeding with this PR as it is now. If still would like to talk about this topic and we later can also apply changes still. So, please reply when you have time.

PR #228, opened 2024-11-19
SD-1775: Issues with parsing the provided public key

nt-gt:
The real issue was the header used `RSA `, when it should not have. However, I also changed the code to use Bouncycastle's PEM parser and writer to delegate some of the code to a library.


preetamnpr (commented):
Looks fine..

PR #234, opened 2024-11-26
Generic code improvements

jkosternl:
* Add useful Javadoc -> please add suggestions if this can be even better, or improve it in your next PR 😉 
* Remove redundant code
* Remove double logging of partyInput
* Fix comparison of incompatible types
* Fix Python method naming in TnT
* Apply `@ParameterizedTest` pattern on TnT

### DIFF
--- a/adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/FilterParameter.java
+++ b/adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/FilterParameter.java
@@ -1,9 +1,5 @@
 package org.dcsa.conformance.standards.adoption.party;
 
-import java.util.Arrays;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 import lombok.Getter;
 
 @Getter
@@ -17,9 +13,4 @@ public enum FilterParameter {
     this.queryParamName = queryParamName;
   }
 
-  public static final Map<String, FilterParameter> byQueryParamName =
-      Arrays.stream(values())
-          .collect(
-              Collectors.toUnmodifiableMap(
-                  FilterParameter::getQueryParamName, Function.identity()));
 }
--- a/core/src/main/java/org/dcsa/conformance/core/toolkit/JsonToolkit.java
+++ b/core/src/main/java/org/dcsa/conformance/core/toolkit/JsonToolkit.java
@@ -42,7 +42,7 @@ public class JsonToolkit {
     }
     if (replacements != null)
       replacements.forEach(
-          (key, value) -> jsonString.set(jsonString.get().replaceAll(key, value)));
+          (key, value) -> jsonString.set(jsonString.get().replace(key, value)));
     return OBJECT_MAPPER.readTree(jsonString.get());
   }
 
--- a/sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
+++ b/sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
@@ -231,27 +231,22 @@ public class ConformanceOrchestrator implements StatefulEntity {
   public void handlePartyInput(JsonNode partyInput) {
     log.info("ConformanceOrchestrator.handlePartyInput(%s)".formatted(partyInput.toPrettyString()));
     if (currentScenarioId == null) {
-      log.info(
-          "Ignoring party input %s: no scenario is currently active"
-              .formatted(partyInput.toPrettyString()));
+      log.info("Ignoring party input: no scenario is currently active");
       return;
     }
 
     ConformanceScenario currentScenario = _getCurrentScenario();
     ConformanceAction nextAction = currentScenario.peekNextAction();
     if (nextAction == null) {
-      log.info(
-          "Ignoring party input %s: the active scenario has no next action"
-              .formatted(partyInput.toPrettyString()));
+      log.info("Ignoring party input: the active scenario has no next action");
       return;
     }
 
     String actionId = partyInput.get("actionId").asText();
     if (!Objects.equals(actionId, nextAction.getId().toString())) {
       log.info(
-          "Ignoring party input %s: the expected next action id is %s in current scenario %s"
-              .formatted(
-                  partyInput.toPrettyString(), nextAction.getId(), currentScenario.toString()));
+          "Ignoring party input: the expected next action id is %s in current scenario %s"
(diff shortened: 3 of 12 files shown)


core/src/main/java/org/dcsa/conformance/core/toolkit/JsonToolkit.java
```diff
@@ -42,7 +42,7 @@ public static JsonNode templateFileToJsonNode(
     }
     if (replacements != null)
       replacements.forEach(
-          (key, value) -> jsonString.set(jsonString.get().replaceAll(key, value)));
+          (key, value) -> jsonString.set(jsonString.get().replace(key, value)));
```
jkosternl, line 45: This change can be discussed, but I did not see that we're using regex to replace data, if so, I'll revert it.

jkosternl: Yes, I can remove it, when we do like to support regex in the key searching. Because the current implementation with `replaceAll` supports regex and `replace` does not.

jkosternl: Ok, I'll revert this change, so we can use regex for search & replace key/values inside sample data.

tnt/src/test/java/org/dcsa/conformance/standards/tnt/checks/TntChecksTest.java
```diff
@@ -1,57 +1,61 @@
...
 
-  @Test
-  void testValidateEventCreatedDateTime() {
-    _check(TODAY, EVENT_CREATED_DATE_TIME_LT, TODAY, false);
-    _check(TODAY, EVENT_CREATED_DATE_TIME_LTE, TODAY, true);
-    _check(TODAY, EVENT_CREATED_DATE_TIME, TODAY, true);
-    _check(TODAY, EVENT_CREATED_DATE_TIME_EQ, TODAY, true);
-    _check(TODAY, EVENT_CREATED_DATE_TIME_GTE, TODAY, true);
-    _check(TODAY, EVENT_CREATED_DATE_TIME_GT, TODAY, false);
+  private static Stream<Arguments> testValidateEventCreatedDateTime() {
```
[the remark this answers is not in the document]
jkosternl: Mostly it is clear if they match, to show that they belong together. But I'll rename this one; fine too.

adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/FilterParameter.java
```diff
@@ -17,9 +13,4 @@ public enum FilterParameter {
     this.queryParamName = queryParamName;
   }
 
-  public static final Map<String, FilterParameter> byQueryParamName =
```
palatsangeetha, line 20: Why this has been removed?

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
```diff
@@ -231,27 +231,22 @@ public JsonNode handleGetPartyPrompt(String partyName) {
   public void handlePartyInput(JsonNode partyInput) {
     log.info("ConformanceOrchestrator.handlePartyInput(%s)".formatted(partyInput.toPrettyString()));
     if (currentScenarioId == null) {
-      log.info(
-          "Ignoring party input %s: no scenario is currently active"
-              .formatted(partyInput.toPrettyString()));
+      log.info("Ignoring party input: no scenario is currently active");
```
palatsangeetha, line 234: Why don't we log the party Input now? Has it been logged anywhere else already?

tnt/src/main/java/org/dcsa/conformance/standards/tnt/action/TntAction.java
```diff
@@ -81,7 +80,7 @@ private void updateCursorFromResponsePayload(ConformanceExchange exchange) {
       }
     }
 
-    if (!dsp.equals(updatedDsp)) dsp.set(updatedDsp);
+    if (!dsp.get().equals(updatedDsp)) dsp.set(updatedDsp);
```
palatsangeetha, line 83: I think this change can be changed in other standards as well as I believe dsp.equals(updatedDsp) is what being done in other standards too.

jkosternl: Yes, my IntelliJ gave this warning. Not sure if it was generated by SonarLint plugin, which I recommend to install, or if it was by IntelliJ itself. Perhaps update your IDE, to fetch the latest enhancements, but this feature was already in for a longer period. 

I will search this in other code as well, and apply the fix. I think I already did at some places in the past PRs btw.

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
jkosternl: Yes indeed. No problem GJ, that happens over time, when software evolves. 

@palatsangeetha, did you see the description of this PR? I put a note in there, to help you guys review the PR: "Remove double logging of partyInput". No issues if you missed it, but just to let you know that it might help you in the future as well.

adoption/src/main/java/org/dcsa/conformance/standards/adoption/party/FilterParameter.java
jkosternl: Because it is not needed and used. Not used code should be cleaned, unless there are very good reasons not to (e.g. going to use it soon).

core/src/main/java/org/dcsa/conformance/core/toolkit/JsonToolkit.java
jkosternl: Then I'll make it `replace`, so regex is not taken into account, if I understood you correctly. I also thought it might cause side effects when user input also contains regex. 

Yeah, your confusing happens quite often by many people, because if feels like `replace` is only doing 1 replacement and `replaceAll` is taking multiple hits into consideration.. But Java is not going to change that, because of backwards compatibility issues..

preetamnpr (approved):

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
jkosternl: No problem. Thanks for checking!

palatsangeetha (approved):

PR #235, opened 2024-11-28
SD-1823 revisit of the Ovs checks with unit test cases.

preetamnpr:


ovs/src/main/java/org/dcsa/conformance/standards/ovs/party/OvsPublisher.java
```diff
@@ -44,6 +44,11 @@ public OvsPublisher(
         orchestratorAuthHeader);
   }
 
+  private static final boolean RETURN_EMPTY_RESPONSE = false;
+  private static final boolean USE_WRONG_ATTRIBUTE_VALUES = false;
+  private static final boolean USE_WRONG_DATE_TIMES = false;
+  private static final boolean USE_WRONG_RESPONSE_STRUCTURE = false;
```
jkosternl, line 50: @preetamnpr, I don't understand the idea behind this structure. This is always false in this class, why are we then using it in the logic, to check on it's value? 
And on line 139 and next: you don't need `else` when the previous statement already returns a value. That prevents not necessary indentation and helps readability.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -30,124 +33,70 @@ public static ActionCheck responseContentChecks(
...
         JsonAttribute.customValidator(
-            "Validate facilitySMDGCode exists and "
-                + "matches in JSON response if request parameter has facilitySMDGCode",
-            body ->
-                validateParameter(
-                    body,
-                    sspSupplier,
-                    OvsFilterParameter.FACILITY_SMDG_CODE,
-                    "*/vesselSchedules/*/transportCalls/*/location/facilitySMDGCode")));
+            "If present, at least schedule attribute must match the corresponding query parameters",
```
palatsangeetha, line 52: Does this mean at least one schedule? I didn't get it from this message string

ovs/src/main/java/org/dcsa/conformance/standards/ovs/party/OvsFilterParameter.java
```diff
@@ -4,22 +4,34 @@
...
-  END_DATE("endDate"),
-  LIMIT("limit"),
+  VESSEL_IMO_NUMBER("vesselIMONumber", false, "*/vesselSchedules/*/vesselIMONumber"),
+  VESSEL_NAME("vesselName", false, "*/vesselSchedules/*/vesselName"),
+  CARRIER_VOYAGE_NUMBER(
+      "carrierVoyageNumber",
+      false,
+      "*/vesselSchedules/*/transportCalls/*/carrierExportVoyageNumber",
+      "*/vesselSchedules/*/transportCalls/*/carrierImportVoyageNumber"),
+  UNIVERSAL_VOYAGE_REFERENCE(
```
palatsangeetha, line 23: Since the USR can be either importVoyageReference or exportVoyageReference, how is this handled in the param matching validation? If one is matching and another not does it throw validation?

ovs/src/main/java/org/dcsa/conformance/standards/ovs/party/OvsPublisher.java
```diff
@@ -130,7 +136,37 @@ public ConformanceResponse handleRequest(ConformanceRequest request) {
 
     Map<String, Collection<String>> headers =
       new HashMap<>(Map.of(API_VERSION, List.of(apiVersion)));
-    return request.createResponse(200, headers, new ConformanceMessageBody(filteredArray));
+
+    if (RETURN_EMPTY_RESPONSE) {
+      return request.createResponse(
+        200, headers, new ConformanceMessageBody(OBJECT_MAPPER.createArrayNode()));
+    } else if (USE_WRONG_ATTRIBUTE_VALUES) {
```
palatsangeetha, line 143: Do we need two different responses for both wrong attributes and dates?

ovs/src/main/java/org/dcsa/conformance/standards/ovs/party/OvsFilterParameter.java
preetamnpr: The code gets the list of carrierExportVoyageNumber , carrierImportVoyageNumber values from the response and tries to match with the query parameter value, if it didnt match then non conformat else its conformant.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
preetamnpr: it should have been "at least one schedule". it corrrect it

ovs/src/test/java/org/dcsa/conformance/standards/ovs/OvsChecksTest.java
```diff
@@ -0,0 +1,478 @@
...
+
+    ActionCheck actionCheck =
+        OvsChecks.responseContentChecks(matchedId, standardVersion, sspSupplier);
+    ((JsonAttributeBasedCheck) actionCheck)
+        .getValidators()
+        .forEach(
+            validator -> {
+              issues.addAll(validator.validate(serviceNodes));
+            });
+    return issues;
```
jkosternl, line 90: If the checks are exposed on package level, you can remove this `executeResponseChecks` method.

jkosternl (changes requested):
I've investigated the code locally and I think you can make it easier, with a few adjustments. Let's have a Team call if you think it is not feasible.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -255,73 +207,116 @@ private Set<String> validateDate(
...
+              .anyMatch(
+                  node ->
+                      !node.getValue().isMissingNode()
+                          && node.getValue().isArray()
+                          && !node.getValue().isEmpty());
+      if (!hasVesselSchedules) {
+        validationErrors.add("Response doesn't have schedules.");
+      }
     }
+    return validationErrors;
```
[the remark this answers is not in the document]
preetamnpr: The return validationErrors looks fine to me, as I have initialized it to add any issues if encountered in the condition.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -255,73 +207,116 @@ private Set<String> validateDate(
...
 
-  private Set<String> validateUniqueTransportCallReference(JsonNode body) {
-    Set<String> transportCallReferences = new HashSet<>();
-    Set<String> errors = new HashSet<>();
+  private static LocalDate stringToISODateTime(String dateTimeString) {
+    try {
+      return OffsetDateTime.parse(dateTimeString, DateTimeFormatter.ISO_DATE_TIME).toLocalDate();
+    } catch (DateTimeParseException e) {
+      log.error("Failed to parse date time string: {}", dateTimeString, e);
+      return null;
```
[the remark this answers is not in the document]
preetamnpr: Here we dont care for the invalid format, hence return null, we just consider only the valid eventDateTime.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -163,84 +112,87 @@ public static ActionCheck responseContentChecks(
...
-                      "The number of schedules exceeds the limit parameter: " + expectedLimit);
+                      "The number of service schedules exceeds the limit parameter: "
+                          + expectedLimit);
                 }
               }
-
               return Set.of();
             }));
 
     return JsonAttribute.contentChecks(
```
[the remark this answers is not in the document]
jkosternl: It was just a suggestion on how you can more easily unit test it. The last sentence on the wrapper is just a side note.. just keep it if it is needed. No worries.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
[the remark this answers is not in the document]
jkosternl: In that case, the `log.error` also does not have value, and you can remove it.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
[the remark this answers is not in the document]
jkosternl: You can just do this: `return Set.of("Response doesn't have schedules.")` at line 317 and similar on 317 instead.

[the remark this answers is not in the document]
preetamnpr: Response body is missing or null. and Response doesn't have schedules. are two different messages.. I think its a matter of perspective.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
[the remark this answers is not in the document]
preetamnpr: I have put this, incase for any investigation, because we have encountered unusual matter. Hence the log.error message.

ovs/src/test/java/org/dcsa/conformance/standards/ovs/OvsChecksTest.java
preetamnpr: U still need the executeReponseChecks because you have the execute the validators. Hence it still remains in the unit test

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -13,234 +14,185 @@
...
+        body -> {
+          Set<String> validationErrors = new LinkedHashSet<>();
+          checkServiceSchedulesExist(body)
+            .forEach(
+              validationError ->
+                validationErrors.add(
+                  "CheckServiceSchedules failed: %s".formatted(validationError)));
+          if (validationErrors.isEmpty()) {
+            return Set.of();
+          }
```
jkosternl, line 43: Line 41-43 can be removed, because line 44 does the same (return empty Set, or with entries).

ovs/src/main/java/org/dcsa/conformance/standards/ovs/checks/OvsChecks.java
```diff
@@ -255,73 +207,116 @@ private Set<String> validateDate(
...
+        return Stream.of();
+      }
     }
   }
+
+  public Set<String> checkServiceSchedulesExist(JsonNode body) {
+
+    if (body == null || body.isMissingNode() || body.isNull()) {
+      return Set.of("Response body is missing or null.");
+    } else {
```
jkosternl, line 308:
```suggestion
    }
```

jkosternl (approved):
Thanks Preetam. Please wait before merging, on the others.

ovs/src/main/java/org/dcsa/conformance/standards/ovs/party/OvsFilterParameter.java
palatsangeetha: Does this validates carrierImportVoyage and exportVoyage or universalImportVoyageNumber and universalExportVoyageNumber? As the standard says 
image

Also if universalImportVoyageNumber is matched and universalExportVoyageNumber is not matched I think that is a valid case as per the standard. How does this validation catches it?

preetamnpr: 1) If the query parameter contains universalVoyageReference, then the jsonPaths from the OvsFilterParameter is fetched, in this case universalImport/ExportVovageParameter values from the response. These values are matched against the query parameter value. If none is matched then non conformant else conformant.  The same applies to CarrierVoyageNumber as well.
2) We are trying to find the matching values in either of the attribute values. If the query parameter is matched in one attribute value and not in another then its a valid case.

palatsangeetha: Thanks Preetam for the clarification

PR #267, opened 2025-01-30
SD-1973 Fix UC3 issue by changing update logic

palatsangeetha:


jkosternl (approved):
I don't fully understand why the `.toUpperCase()` is needed, but I guess that's because of the schema requirements and therefor approving.

PR #270, opened 2025-02-06
SD-1909-BKG update the action prompts to show non null values

palatsangeetha:


jkosternl (changes requested):
Good improvements! I was almost approving it, but saw the method logic duplicated twice, so now the same logic is in 3 classes. Can you put the method somewhere that you can reuse it? Public is fine.

PR #277, opened 2025-02-19
SD-2029 JIT 2.0: Add Inline Instructions

jkosternl:
* Rename serviceType to serviceTypeCode (forgot some instances before)
* Remove unused imports

FYI: the build is failing because of duplicated lines, but in fact, they are not duplicated, but pointing to different Markdown files, which SonarQube seems to skip.


jit/src/main/resources/standards/jit/instructions/prompt-process-oob-input.md
```diff
@@ -0,0 +1,2 @@
+Supply the out-of-band %s timestamp that you (could) have received out-of-band from the service provider, using the
+format below:
```
palatsangeetha, line 2: Does it means the format in the text box?

jkosternl: Thanks both. I'll change it a bit, so it gets clear.

palatsangeetha (approved):

PR #284, opened 2025-03-03
SD-2038-Correct CS checks

palatsangeetha:


jkosternl (approved):
I see quite a lot of checks removed, because they're not backed by the standard? Great that you've added unit tests directly as well 👍🏻

palatsangeetha:
> I see quite a lot of checks removed, because they're not backed by the standard? Great that you've added unit tests directly as well 👍🏻

Yeah had to remove those checks as we have decided not to check filter parameters are present in the payload as well as all the checks the are not explicitly stated in the standard. Thanks for the Approval.

PR #287, opened 2025-03-05
SD-1560 Reduce AWS Lambda deployment size with 62%

jkosternl:
FYI: 
- Jar file goes from 31MB to 22MB. 
- Tested it in AWS deployment (using CDK) and fixed some errors in few trails, hence the added inclusion filters in the `pom.xml`.


jkosternl:
> This seems to be unrelated with what's actually requested in SD-1560

Yes, @gj0dcsa, I am still typing in the Jira issue to explain it. 😄

PR #291, opened 2025-03-12
SD-2053 Upgrade Web UI libraries, fixing esbuild and prismjs vulnerabilities

jkosternl:
Finally, it is possible to fix `esbuild` 😄 also fixes `prismjs`. 
After this PR: `found 0 vulnerabilities`

### DIFF
--- a/webui/package-lock.json
+++ b/webui/package-lock.json
@@ -56,13 +56,13 @@
       }
     },
     "node_modules/@angular-devkit/architect": {
-      "version": "0.1902.0",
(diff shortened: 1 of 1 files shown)


palatsangeetha (approved):

jkosternl:
@gj0dcsa, I applied a second upgrade today, which fixes all new vulnerabilities. Looking at the diff, I don't see downgraded libs. So I think this one should be okay to merge. Waiting for your approval to do so.

PR #293, opened 2025-03-18
SD-2076 JIT schema generation with Swagger Core lib

jkosternl:


jit/src/main/java/org/dcsa/conformance/standards/jit/schema/SchemaParams.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String API_VERSION_MAJOR_REF = "#/components/parameters/Api-Version-Major";
+
+  public static Parameter getPortCallIDPathParam() {
+    return new Parameter()
+        .name("portCallID")
+        .in("path")
+        .required(true)
+        .description("The provider created identifier for the **Port Call**.")
+        .schema(getUUIDSchema());
+  }
```
HenrikHL, line 21: Personally I would see this as a public static method on a `PortCallID` class (which should also be used in the `PortCall` object. This way the definition for the PortCallID and the PathParam follow each other. In the YAML spec - this was always a problem for me to align the two...

jit/src/main/java/org/dcsa/conformance/standards/jit/schema/Vessel.java
```diff
@@ -0,0 +1,107 @@
...
+    - `CARC` (Car carrier)
+    - `PASS` (Passenger)
+    - `FERY` (Ferry)
+    - `BULK` (Bulk)
+    - `TANK` (Tanker)
+    - `LGTK` (Liquefied gas tanker)
+    - `ASSI` (Assistance)
+    - `PILO` (Pilot boat)
+""",
+      example = "CONT")
```
HenrikHL, line 105: Can this annotation be located in the `VesselType` instead of here? (As it would be the same description in case `vesselType` is used in other objects...

HenrikHL: This is important to know if the above is possible - same goes if I want the example to be modified compared to what is provided in the base object (but still keep all other annotation-attributes)

jkosternl: As discussed in Teams: it is possible, I just moved it to `VesselType` enum class, but you can not override the description field in the `Vessel` class unfortunately. It keeps using the one in `VesselType` enum.

PR #316, opened 2025-05-22
SD-2010-Fix for missing notification support

palatsangeetha:


palatsangeetha:
> The reason to call it /dev/null was that everyone would immediately understand what it's for, but /dev/empty will also work I guess 🙂

Thanks for approving and the comment @gj0dcsa . I will change it to dev/null so that it's more obvious what it's for.

PR #503, opened 2026-04-09
Sd-1897-url validation updated

palatsangeetha:


core/src/main/java/org/dcsa/conformance/core/toolkit/Url.java
```diff
@@ -0,0 +1,129 @@
...
+   * Returns true if this URL is empty or blank.
+   */
+  public boolean isBlank() {
+    return value.isBlank();
+  }
+
+  /**
+   * Returns the length of the URL string.
+   */
+  public int length() {
```
pedrocarvalhodcsa, line 105: is this method used?

core/src/main/java/org/dcsa/conformance/core/toolkit/Url.java
```diff
@@ -0,0 +1,129 @@
...
+    return value.length();
+  }
+
+
+  @Override
+  public String toString() {
```
pedrocarvalhodcsa, line 111: Do we need the toString, equals and hashCode?

pedrocarvalhodcsa (commented):
All seems good. Just take a look at the comments from Copilot and the other AI bot to see if they make sense and they need changes

core/src/main/java/org/dcsa/conformance/core/toolkit/Url.java
palatsangeetha: toSting and length() is being used, toString indeirectly through lombok. The other two are also added for future if url is being used in collections.

pedrocarvalhodcsa (approved):

PR #531, opened 2026-07-24
SD-3171: Validate incoming requests to the sandbox

pedrocarvalhodcsa:


sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
```diff
@@ -312,6 +312,24 @@ public void handlePartyTrafficExchange(ConformanceExchange exchange) {
...
+
+    ConformanceScenario currentScenario = _getCurrentScenario();
+    ConformanceAction nextAction = currentScenario.peekNextAction();
+
+    if (nextAction == null || nextAction.getTargetPartyName() == null) {
+      return "The party '%s' is not expected to receive an exchange at this time. Read the sandbox's instructions and proceed with the scenario."
+        .formatted(partyName);
+    }
+
+    return null;
```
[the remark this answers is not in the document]
pedrocarvalhodcsa: But some of our actions have optional requests to the party that its no the main target party (e.g. notifications), therefore we can't restrict it

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
```diff
@@ -312,6 +312,24 @@ public void handlePartyTrafficExchange(ConformanceExchange exchange) {
...
+    if (currentScenarioId == null) {
+      return "There is no active scenario. This party should not be receiving any requests at this time.";
+    }
+
+    ConformanceScenario currentScenario = _getCurrentScenario();
+    ConformanceAction nextAction = currentScenario.peekNextAction();
+
+    if (nextAction == null || nextAction.getTargetPartyName() == null) {
+      return "The party '%s' is not expected to receive an exchange at this time. Read the sandbox's instructions and proceed with the scenario."
+        .formatted(partyName);
```
gj0dcsa, line 327: "API requests to party '%s' are only allowed when a scenario is running and the current action allows the exchange of API requests."

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
```diff
@@ -312,6 +312,24 @@ public void handlePartyTrafficExchange(ConformanceExchange exchange) {
...
+    log.info("ConformanceOrchestrator.validatePartyCanReceiveExchange(%s)".formatted(partyName));
+
+    if (currentScenarioId == null) {
+      return "There is no active scenario. This party should not be receiving any requests at this time.";
+    }
+
+    ConformanceScenario currentScenario = _getCurrentScenario();
+    ConformanceAction nextAction = currentScenario.peekNextAction();
+
+    if (nextAction == null || nextAction.getTargetPartyName() == null) {
```
gj0dcsa, line 325: I don't think it's that simple: this doesn't address the support situation that has triggered this ticket, in which the adopter was trying to exchange API requests during the "supply scenario parameters" action 🤔

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
```diff
@@ -312,6 +312,24 @@ public void handlePartyTrafficExchange(ConformanceExchange exchange) {
     }
   }
 
+  public String validatePartyCanReceiveExchange(String partyName) {
+    log.info("ConformanceOrchestrator.validatePartyCanReceiveExchange(%s)".formatted(partyName));
+
+    if (currentScenarioId == null) {
+      return "There is no active scenario. This party should not be receiving any requests at this time.";
```
gj0dcsa: "API requests to party '%s' are only allowed when a scenario is running and the current action allows the exchange of API requests."

gj0dcsa (changes requested):

sandbox/src/main/java/org/dcsa/conformance/sandbox/ConformanceOrchestrator.java
pedrocarvalhodcsa: It addresses, I debugged it and when you are in the SSP action the nextAction.getTargetPartyName() is null. 
I also tried adding the condition when `!partyName. equals(nextAction.getTargetPartyName())` but this is not doable because there are some scenarios that optionally allow both sides to make requests (e.g. booking and ebl notifications)

gj0dcsa (approved):

